PR #10, opened 2026-05-03
Drop unnecessary shift in traj_append_step

bobbyhwilder:
Trivial cleanup — `local traj_dir="$1"; shift; local json="$1"` is equivalent to `local traj_dir="$1"; local json="$2"`, just with extra steps. No behavior change.


bobbyhwilder:
Superseded by upstream commit e63434f (same change, applied directly rather than via merge).

PR #12, opened 2026-06-15
SHELLM_THINKER_ENV override, bash 3.2 compat, executive infinite loop

bobbyhwilder:
## Summary

Three related fixes:

- **`SHELLM_THINKER_ENV` override** — `_build_shellm_flags` now emits `--env` and `--workdir`, gated by `SHELLM_THINKER_ENV` (defaults to `$IDENTITY_NAME`). All 8 step scripts drop their hardcoded `--env`/`--workdir`. Usage: `SHELLM_THINKER_ENV=local thinkers start` skips Docker envs.

- **bash 3.2 compatibility** — macOS ships bash 3.2; two bash-4+ constructs were silently crashing:
  - `actor/step`: `${is_reply^^}` → portable `tr` uppercase. **This was the root cause of zero actions/forks in the trajectory** — every actor invocation died at line 59 before reaching the shellm call.
  - `shellm-docker-broker`: `local -A` (associative array) → indexed array deduped via `sort -u`.

- **Executive infinite-loop guarantee** — the executive relies on `trigger_self=true` to re-fire after each thought, but two conditions broke the loop: empty LLM response (`exit 1`, no traj append) and duplicate response (dedup guard skips append). Both leave the FIFO dry. Fix: always append — placeholder on empty (with 1s backoff), `" (repeat)"` suffix on duplicate.

## Test plan

- [x] `bash -n` passes on all 11 files under macOS bash 3.2
- [x] `shellm-docker-broker` `symlink_mounts` rewrite verified under bash 3.2 with unit tests (single link, multiple links with dedup + dominance filtering, empty/missing workdir)
- [x] Actor bash-4 crash confirmed in `actor.log` (52 identical `bad substitution` lines); fix verified by syntax check
- [ ] Restart thinkers and confirm executive no longer goes idle + actor forks appear in trajectory

### DIFF
--- a/bin/shellm-docker-broker
+++ b/bin/shellm-docker-broker
@@ -69,34 +69,39 @@ path_under() {
 
 symlink_mounts() {
     local workdir="$1"
-    local -A mount_dirs=()
-    while IFS= read -r -d '' link; do
-        local target
-        target=$(readlink -f "$link" 2>/dev/null) || continue
-        path_under "$target" "$workdir" && continue
-        [[ -e "$target" ]] || continue
-        local parent
-        parent=$(dirname "$target")
-        mount_dirs["$parent"]=1
-    done < <(find "$workdir" -type l -print0 2>/dev/null)
+
+    # bash 3.2 lacks associative arrays, so dedupe via sort -u into an
+    # indexed array.  Paths containing newlines are not supported (the old
+    # implementation also could not survive them).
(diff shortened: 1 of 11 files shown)


andyk:
Closing in favor of targeted PRs for the remaining bug fixes. The actor bash 3.2 compat and hardcoded --env/--workdir removal were already merged via PR #14.

PR #18, opened 2026-07-10
Fix actor chat replies, add OpenRouter support, guard system-dir mounts

andyk:
## Summary

Three fixes found while debugging why bobby's actor stopped replying in chat, plus OpenRouter model support.

### Fix: actor unable to reply in chat (`thinkers/_lib/common.sh`)
`chat reply` hard-requires `IDENTITY_NAME`, but `_build_shellm_flags` never passed it into the thinker's shellm env. Inside the container the actor got `chat: error: IDENTITY_NAME not set`, then flailed into `chat send` variants that also failed (`--from`/`--to` errors). Replies only ever landed when the model happened to improvise an inline `IDENTITY_NAME=bobby` prefix on its own. Now passed via `--var` (non-directory values are plain env vars — no mount/metadata changes, existing envs stay compatible).

### Feature: OpenRouter provider in `llm`
- Vendor-prefixed model names (`z-ai/glm-5.2`, `meta-llama/...` — anything with a `/`) auto-route to OpenRouter; `--provider openrouter` works explicitly.
- OpenAI-compatible wire format: payload builder, extractor, and SSE streamer are aliases of the OpenAI handlers (including `--thinking` → `reasoning.effort`).
- `OPENROUTER_API_KEY` auth, `LLM_API_URL` override respected, 16384 default max tokens for vendor-prefixed models.
- Verified: streaming, non-streaming, missing-key error path, and a full `shellm -m z-ai/glm-5.2` run end-to-end in Docker.

### Fix: `llm` crash on streaming curl errors
`local` used outside a function in the stream error path — any curl failure crashed llm with `local: can only be used in a function`, masking the real error.

### Fix: shellm mounted macOS `/usr/bin` over the container's (`bin/shellm`)
The symlink-target mount scanner followed a stale venv's `python3 → /usr/bin/python3` link and bind-mounted the host's `/usr/bin` over Ubuntu's, breaking every exec in newly created containers (`"sleep": executable file not found in $PATH`). Host system directories (`/usr`, `/bin`, `/etc`, `/var`, `/System`, `/Library`, ...) are now excluded from symlink-target mounting.

### DIFF
--- a/bin/llm
+++ b/bin/llm
@@ -54,7 +54,7 @@ Options:
   -t, --max-tokens N       Max output tokens. Default: model-specific cap
   -M, --messages JSON      Messages array JSON, e.g. [{"role":"user","content":"hi"}]
   --no-stream              Disable streaming (streaming is on by default)
-  --provider NAME          Provider: anthropic, openai, or gemini
+  --provider NAME          Provider: anthropic, openai, gemini, or openrouter
   --thinking [LEVEL]       Enable thinking (Anthropic: adaptive; Gemini: minimal/low/medium/high)
   --effort LEVEL           Thinking effort/provider output effort
   --raw                    Print raw non-streaming API response JSON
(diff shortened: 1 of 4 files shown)


nickjalbert:
cool!  the identity name one looks very familiar from my testing yesterday

PR #24, opened 2026-07-28
Add monolith thinker: the whole mind in one solo router

bobbyhwilder:
## Summary
Introduces `monolith`, a single thinker that runs solo and does the whole job of the old roster (inner_monologue, actor, learning, mind_wanderer, goals_manager, values_manager) by **routing to one function per wakeup**. Opt-in: install only `thinkers/monolith/` into an identity.

Each wakeup takes one of two paths:

- **Fast reply** — a chat `message` addressed to us is answered *immediately*: one `llm` call composed from the recent stream, then `chat reply`. No thinking step, no tools, no agentic loop — lowest latency the single process allows. A bookkeeping `observation` records the reply and keeps the loop alive. Work a message needs happens later via the `act` route.
- **Router** — for any other trigger, one `shellm` agentic run over a function menu (`act / think / learn / recall / goals / values / idle`); the model picks one and carries it out, writing steps via `traj append`.

## Design notes
- **Shared root trajectory** — writes via `traj append` to `TRAJ_ID`; runs pass `--traj "$TRAJ_ID"`. One source of truth.
- **Self-loop guard** — `chat` stamps `source:"chat"` on outgoing replies too, so we only treat a `message` as a trigger when `to == IDENTITY_NAME`.
- **Idle backoff** — when re-fired by our own idle step, sleep with doubling backoff (5s→60s); any real trigger resets it.
- **Liveness net** — the loop is perpetual only if each wakeup appends a *subscribed-type* step. `shellm` always appends its own `prompt`/`shellm-run` machinery (not subscribed), so the net tracks the last subscribed-type step id and appends a fallback `idle` if a run left none — otherwise `trigger_self` would silently go dark.

## Testing
Drove every path against throwaway scratch identities and then bobby's live identity: fast reply (real `chat reply`), successful router run (exactly one model step, no double-append), failed run (one fallback idle → loop survives), self-loop guard (no append), idle-backoff math (5→10→20→40→60). Running solo in an identity, it routes to `idle` and backs off when the mind has nothing pressing.

See `design/monolith_thinker.md` for the full architecture and the earlier two-thinker draft it supersedes.

### DIFF
--- a/design/monolith_thinker.md
+++ b/design/monolith_thinker.md
@@ -1,0 +1,154 @@
+# Monolith thinker
+
+Status: draft
+Replaces (when enabled): the multi-thinker roster (inner_monologue, actor,
+goals_manager, learning, mind_wanderer, values_manager) with a single thinker,
+`monolith`, that runs solo and handles every function — including chat.
+
+## Motivation
+
+The current roster runs six thinkers that mostly share one shape: wake on new
+steps, read the recent stream, decide whether to do their one job, maybe write
+a step. Coordination between them is emergent and occasionally pathological
+(double replies, mimicked conventions, loops).
+
+But most thinker "jobs" are alternatives, not parallel processes. At any given
+wakeup the mind should mostly do ONE of: advance the monologue, execute an
+action, reply to a message, store a lesson, recall a memory, tend goals/values,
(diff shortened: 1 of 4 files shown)


nickjalbert:
🚀

PR #29, opened 2026-08-14
Roster to monolith + responder: add responder thinker, remove the legacy six

bobbyhwilder:
**Draft.** Moves the thinker roster to **monolith + responder** (+ `_lib`).

## 1. Add the responder thinker (first draft)
A dedicated, minimal **chat-reply thinker**: subscribes to `message` only, and on an inbound message replies with a single `llm` call — no shellm agentic loop, no tools — for the lowest possible reply latency. The "responder" half of the two-thinker split (`monolith` router for cognition + `responder` for chat) that `monolith_thinker.md` predicted as the escape hatch if chat latency ever needed isolating.

`thinkers/responder/step` does: reply guards (to-us only, self-message drop), an idempotency scan, a `reply_claim` step staked before composing, a one-`llm`-call reply with `NO_REPLY` support, `reply_to` stamping, and `decision:`-tagged observations. `design/responder_thinker.md` is the reverse-engineered design note.

**Known gap (documented, not fixed):** the `reply_claim` coordination is one-sided — the monolith does not check `reply_claim` (also absent from `trajectory_spec.md`), so monolith + responder can still race to double-reply until the monolith honors the claim or its own fast-reply path is disabled. See the design note.

## 2. Remove the legacy multi-thinker roster
Delete `actor`, `goals_manager`, `inner_monologue`, `learning`, `mind_wanderer`, `values_manager` — the six thinkers the monolith collapses into one. After this, the bundled roster is `monolith` + `responder` + `_lib`, so `identity new` gives new minds only those. Existing identities keep whatever they installed; their symlinks to removed thinkers go stale and can be pruned.


bobbyhwilder:
Superseded by #31, which consolidates this with the other two related PRs into one branch. Closing in favor of the single PR.

PR #30, opened 2026-08-14
Verify shellm tools aren't shadowed by system binaries on PATH

bobbyhwilder:
## Why
macOS ships `/usr/sbin/chat` (the PPP tool) and `/usr/bin/view` (vim). When a system dir precedes the shellm bin dir on PATH, `chat`/`view` silently resolve to the *wrong* program. On 2026-08-14 this made every monolith/responder `chat reply` fail silently — bobby *saw* the messages (the responder fired) but no reply ever went out, because `chat reply` was invoking `/usr/sbin/chat`.

Checking PATH *ordering* is fragile; instead, both layers here **resolve each tool and compare it to the copy we expect** (`realpath`).

## Two complementary layers
- **`thinkers` (runtime guard):** at `thinkers start`, prepend our own bin dir (so our tools win over earlier system dirs), then verify each critical tool on PATH `realpath`s to our copy — `exit 1` with a clear error if one is still shadowed. Self-heals the dispatcher and every step process it spawns. *(Verified live: bobby started under the broken PATH order and self-healed — replies delivered.)*
- **`install.sh` (install-time warning):** when `$PREFIX` is already on PATH, verify each installed tool resolves to the copy just installed; **warn** (install succeeded — PATH order is the user's to fix), naming each shadow and how to reorder PATH. This covers contexts the runtime guard can't: your interactive shell, the chat TUI, the web control plane, cron.

## Testing
- `bash -n` clean on both.
- Shadow detection unit-tested: a system-style `chat` earlier in PATH is flagged; correct order passes.
- The `thinkers` guard exercised end-to-end against the real `/usr/sbin/chat` shadow.

### DIFF
--- a/bin/thinkers
+++ b/bin/thinkers
@@ -848,12 +848,45 @@ _append_wake_gap_note() {
         | traj append --traj_dir "$TRAJ_DIR" "${traj_file#"$TRAJ_DIR"/}" >/dev/null 2>&1 || true
 }
 
+# Guard against shadowed tools. The shellm CLIs live together next to this
+# script, so our own resolved location is the canonical bin dir. Prepend it (so
+# our tools beat same-named system binaries — macOS ships /usr/sbin/chat,
+# /usr/bin/view, ...), then VERIFY each critical tool on PATH actually resolves
+# to OUR copy. Fail loudly rather than silently running the wrong program — the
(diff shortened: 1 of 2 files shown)


bobbyhwilder:
Superseded by #31, which consolidates this with the other two related PRs into one branch. Closing in favor of the single PR.

PR #45, opened 2026-08-25
telegram: a malformed API reply no longer ends the poll loop

### LINKED ISSUE #44: Telegram bridge: a non-JSON API reply ends the poll loop
`Bot._call` calls `response.json()` with no guard:

https://github.com/laude-institute/headlong/blob/main/telegram/src/headlong_telegram/api.py#L31-L35

When Telegram or something in front of it answers with a non-JSON body, a 502 HTML page or an empty response, that raises `json.JSONDecodeError`. It subclasses `ValueError`, so it is neither `ApiError` nor `httpx.HTTPError`, and the poll loop catches exactly those two:

https://github.com/laude-institute/headlong/blob/main/telegram/src/headlong_telegram/inbound.py#L60-L69

So the exception walks past the retry branch and out of `Inbound.run`.

Reproduction:

```python
import httpx
from headlong_telegram.api import ApiError, Bot

bot = Bot("t")
bot._client = httpx.Client(transport=httpx.MockTransport(
    lambda r: httpx.Response(502, text="<html>502 Bad Gateway</html>")))

try:
    bot.get_updates(0)
except Exception as e:
    print(type(e).__name__, isinstance(e, (ApiError, httpx.HTTPError)))
# JSONDecodeError False
```

The unit has `Restart=always` with `RestartSec=5`, so the bridge comes back rather than staying down. What it loses is the in-memory `_seen_msg_ids` dedup map, and since delivery is at-least-once, a redelivered update after the restart can double-post to the mind log. A bad minute upstream turns into a restart cycle instead of the retry the loop was written to do.

The module docstring already states the contract this misses: "Every call returns the `result` payload on ok=true and raises ApiError otherwise."

`telegram/tests/` has no coverage for `api.py`, and `test_inbound.py` drives a fake bot, which is why the existing suite does not see it.

I have a fix and will open a PR against this issue.

beardthelion:
Fixes #44

`Bot._call` called `response.json()` unguarded, so a non-JSON body raised `json.JSONDecodeError`. That subclasses `ValueError`, which is neither `ApiError` nor `httpx.HTTPError`, and those are exactly the two the poll loop catches. The exception walked past the retry branch and ended `Inbound.run`. systemd restarts the bridge five seconds later, so what is actually lost is the in-memory `_seen_msg_ids` map, and at-least-once delivery can then double-post a redelivered update.

The module docstring already promised the behavior this restores: a call returns `result` or raises `ApiError`. So `_call` now converts three shapes into `ApiError`: a body that is not JSON, a payload that is not an object (a bare list would have raised `AttributeError`), and an `ok` response with no `result` (previously a `KeyError`). None of those three were caught anywhere, so this only widens what the existing `except ApiError` sites already handle, in `inbound.py` and `outbound.py` both.

The error message truncates the response body to 200 characters and leaves the URL out, since the base URL carries the bot token.

## Tests

`api.py` had no tests, so `telegram/tests/test_api.py` is new: the non-JSON body, a JSON error status, `ok: false` still carrying its description, and the ok path still returning `result`.

The regression test in `test_inbound.py` drives a real `Bot` over `httpx.MockTransport` rather than the fake bot the other tests use. That is the point of it: the failure lives in how `_call` surfaces a transport-shaped error, which a fake bot cannot produce, and it is why the existing suite did not catch this.

Both fail without the fix and pass with it. `uv run --project telegram pytest telegram/tests -q` goes from 30 to 35 passing.

### DIFF
--- a/telegram/src/headlong_telegram/api.py
+++ b/telegram/src/headlong_telegram/api.py
@@ -30,9 +30,24 @@ class Bot:
 
     def _call(self, method: str, **params: Any) -> Any:
         response = self._client.post(f"{self._base}/{method}", json=params)
-        payload = response.json()
+        # Anything that is not a well-formed Telegram envelope has to leave here
+        # as ApiError. The poll loop retries on (ApiError, httpx.HTTPError) and
+        # dies on everything else, so a 502 HTML page from an edge would
+        # otherwise raise JSONDecodeError straight through it. The response text
+        # is truncated and the URL left out because the URL carries the token.
+        try:
+            payload = response.json()
+        except ValueError:
+            raise ApiError(
+                f"{method}: non-JSON response (HTTP {response.status_code}): "
+                f"{response.text[:200]}"
+            ) from None
+        if not isinstance(payload, dict):
+            raise ApiError(f"{method}: unexpected response shape {type(payload).__name__}")
         if not payload.get("ok"):
             raise ApiError(f"{method}: {payload.get('description', response.text)}")
+        if "result" not in payload:
+            raise ApiError(f"{method}: ok response carried no result")
         return payload["result"]
 
     def get_me(self) -> dict[str, Any]:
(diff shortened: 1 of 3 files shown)


nickjalbert:
LGTM!  Thank you for the contribution.

PR #46, opened 2026-08-25
Add optional Codex ChatGPT subscription provider

dhruv-anand-aintech:
## Summary

- add an optional `codex` provider backed by the official `codex app-server` JSONL protocol
- preserve the existing API-key providers and make Codex selection explicit through `SHELLM_PROVIDER=codex`
- stream assistant messages and reasoning summaries, record subscription billing provenance, and fail clearly on unsupported interactive approvals
- add a mocked provider test and document ChatGPT-managed Codex setup

## Why

Headlong currently requires a vendor API key. Codex already supports ChatGPT-managed authentication and token refresh. This adapter uses the supported app-server boundary instead of reading or replaying OAuth tokens or treating a ChatGPT login token as an OpenAI API key.

## Validation

- `tests/test_codex_provider.sh`
- `tests/run-all.sh llm` (70 passed, 0 failed)
- local smoke test with the installed Codex CLI, ChatGPT-managed auth, and `gpt-5.6-luna`

## Notes

The current app-server turn completion event does not expose token counts to this adapter, so usage records identify `billing_source=chatgpt_subscription` and set `usage_available=false` rather than inventing counts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,8 +39,8 @@ life, and opens a dashboard where you can watch its mind run:
 curl -fsSL https://headlong.ai/install.sh | bash
 ```
 
-You'll need bash 3.2+, git, curl, jq, and an LLM API key (Anthropic,
-OpenAI, Gemini, or OpenRouter); the dashboard also needs
+You'll need bash 3.2+, git, curl, jq, and either an LLM API key (Anthropic,
+OpenAI, Gemini, or OpenRouter) or an authenticated Codex CLI installation; the dashboard also needs
 [uv](https://docs.astral.sh/uv/) and bun or node, and the installer offers
 to fetch those. 
 
--- a/bin/llm
+++ b/bin/llm
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
+SCRIPT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)"
+
 # llm — Minimal multi-provider LLM CLI tool
 # Supports the Anthropic, OpenAI, Gemini, and OpenRouter APIs.
 
@@ -146,7 +148,7 @@ Options:
   -t, --max-tokens N       Max output tokens. Default: model-specific cap
   -M, --messages JSON      Messages array JSON, e.g. [{"role":"user","content":"hi"}]
   --no-stream              Disable streaming (streaming is on by default)
-  --provider NAME          Provider: anthropic, openai, gemini, or openrouter
+  --provider NAME          Provider: anthropic, openai, gemini, openrouter, or codex
   --thinking [LEVEL]       Enable thinking (Anthropic: adaptive; Gemini: minimal/low/medium/high)
   --effort LEVEL           Thinking effort/provider output effort
   --raw                    Print raw non-streaming API response JSON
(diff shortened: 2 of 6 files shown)


nickjalbert (changes requested):
From review:

```
1. `tools/codex-app-server-llm`: effort is read from `os.environ["LLM_EFFORT"]`, which is never set; accept it as a validated `--effort` flag instead (the adapter invoker passes one).
2. `tools/codex-app-server-llm`: `stderr=subprocess.DEVNULL` discards every Codex diagnostic (not signed in, unknown subcommand, model unavailable); capture it and include a bounded tail in the `fail()` message. Log unparseable stdout lines rather than `continue`.
3. `tools/codex-app-server-llm`: guard against internal wedging: `turn/start` is written from inside the stdout read loop, so a prompt larger than the pipe buffer can deadlock with stdout undrained — write from a separate thread or after draining. Also guard `proc.wait(timeout=2)` against `TimeoutExpired`. (The outer wall-clock deadline and kill are handled by the adapter invoker.)
4. `tools/codex-app-server-llm`: `thread/start` sends `cwd=os.getcwd()` with no explicit sandbox/approval policy; approvals are only rejected reactively, so a permissive operator Codex config could act on the host cwd outside Headlong's containment boundary. Request the most restrictive available mode explicitly, or pass a throwaway cwd, and document the requirement.
5. `tools/codex-app-server-llm`: the message array is flattened into `[role]\n{content}` blocks in one text item, so untrusted Slack/Telegram text can forge `[assistant]`/`[system]` turn boundaries. Use structured per-role input if the protocol allows, or a delimiter that user text cannot produce.
6. Tests: extend the mock into scenarios covering an `item/*/requestApproval` message, JSON-RPC `error` on ids 1/2/3, a non-`completed` turn status, malformed/non-JSON stdout, and a missing `codex` binary; assert non-zero exit and the stderr text for each.
```

Thanks for building this, and sorry the PR sat while I worked through the provider policy it raised. The review above covers the code changes needed. Here is the decision about where the integration belongs.

Headlong now has a written [provider policy](https://github.com/laude-institute/headlong/blob/main/design/providers.md) and a working adapter interface. `bin/llm` will not gain a Codex specific branch or a Python dependency. Instead, operators can set `LLM_PROVIDER=adapter` and `LLM_ADAPTER=/path/to/executable` to run an external provider adapter.

The adapter receives messages as JSON on stdin. `bin/llm` passes the model, token limit, effort, thinking, streaming, and system prompt through documented flags. The adapter can report token usage through `LLM_USAGE_FILE`. `bin/llm` continues to manage the deadline, child process cleanup, health marker, and usage ledger.

Your `codex-app-server-llm` tool already contains most of the provider logic needed for an adapter. If you are willing to rework the PR, I would like it to become the first adapter built against the new contract. The executable can remain under `tools/` or live in its own repository. The Codex branch and other Codex specific behavior should come out of `bin/llm`.

The adapter should also address the findings from the review, including stderr handling, the effort flag, failure tests, and the sandbox behavior around `thread/start`.

Adapters are trusted local code and receive the caller's environment. Codex authentication, token storage, and token refresh therefore remain inside the adapter, which is the right place for subscription and OAuth based authentication.

Quick edit: One concrete implementation note: please also remove the `SHELLM_PROVIDER` forwarding change from `bin/shellm`. The merged adapter interface uses inherited `LLM_PROVIDER`, so the adapter rework should not require any `bin/shellm` changes

PR #47, opened 2026-08-25
Docker broker: contain compose host paths by resolution, and check the fields it never read

beardthelion:
`compose_validate_model` is the only gate between a compose model the sandbox authors and the host filesystem: `handle_compose` runs the real `docker compose` the moment it returns 0. It was not containing host paths. Both commits are about that, and both are reachable only with `SHELLM_DOCKER_ACCESS=broker`, which is opt-in.

**Containment by resolution, not by string prefix.** `under_allowed_root` was `startswith($workdir)`, and nothing resolved the path first, so three things got through: a sibling directory sharing the prefix (`/host/work` accepted a bind on `/host/work-secret`), an absolute source carrying `..` (compose normalizes a relative source but leaves an absolute one verbatim, checked on 2.40.3), and a symlink created inside the workdir pointing out. Every other containment site in the file already routes through `realpath` and `path_under()`; this one now does too. jq emits the candidate paths, bash rejects a relative path or any `..` component, resolves the rest, and gates with `path_under()`. Resolution is lenient about a missing tail because compose creates a missing bind source, and `realpath -m` is GNU-only while this also runs on macOS.

Two things fell out of that. The `workdir_alias` branch is gone, since an aliased spelling now resolves to the workdir like anything else. And a bind whose source is missing or not a string, or a model the extraction cannot parse, is rejected rather than skipped.

**Every host path the model names.** The check only ever read `services[].volumes[]` binds, build contexts and absolute Dockerfiles. A named volume with `driver_opts.device`, `secrets:`/`configs:` with `file:`, `build.additional_contexts` and `env_file` all name a host path and reach the daemon without passing any of them. `env_file` needs a second `--no-interpolate` render to see at all: the default render reads each file and inlines the values into `environment`, leaving `env_file` null, so the paths are gone before validation runs. `config` is no longer exempt from validation either, because the render it hands back carries those inlined values.

Two fail-closed edges worth flagging. `additional_contexts` values are checked only when absolute, since `docker-image://` and `service:` references share that map. A `driver_opts.device` is always checked, so a device that is not an absolute path (an NFS `host:/export`) is rejected rather than passed through.

## Verification

`tests/test_broker_compose_containment.sh` is new and drives the real function against a stubbed `docker compose config`. 24 cases; every reject asserts rc 65 rather than merely nonzero, and the accept cases and the unrelated rejections are there so a validator that denied everything, or never ran, could not pass. Each case is red against the code it fixes.

Separately, every escape was run end to end against a real daemon, with a brokered container reading a host file outside the workdir:

| | before | after |
|---|---|---|
| sibling prefix, `..`, symlink | host file read | rejected, rc 65 |
| volume `driver_opts.device` | host file read | rejected, rc 65 |
| `secrets:` / `configs:` `file:` | host file read | rejected, rc 65 |
| `env_file` into the container | host file read | rejected, rc 65 |
| `env_file` via `compose config` | rendered to the caller | rejected, rc 65 |

Positive controls in the same setup, all still working: a relative bind plus relative `env_file`, a build from a context inside the workdir, a named volume device inside it, a legitimate mount, and `config` on a clean model.

`tests/run-all.sh` is 22/22 and `shellcheck -S warning` is clean on both files. Not exercised: the facade and transport hop (the tests drive `handle_compose` directly), compose subcommands other than `up`, `config` and `build`, and the macOS bash 3.2 job, which CI covers.

### DIFF
--- a/tests/test_broker_compose_containment.sh
+++ b/tests/test_broker_compose_containment.sh
@@ -2,15 +2,25 @@
 # tests/test_broker_compose_containment.sh — the broker's compose validator
 # must contain host paths at a directory boundary, not a string prefix.
 #
-# compose_validate_model is the ONLY gate on the bind sources, build contexts
-# and absolute Dockerfiles in a compose model the sandbox hands the broker;
-# handle_compose runs the real `docker compose` as soon as it returns 0. A
-# textual prefix test is not containment: it lets /host/work-secret through
-# while the workdir is /host/work, and `docker compose config` hands the
-# validator paths it has NOT normalised (an absolute source keeps its `..`,
+# compose_validate_model is the ONLY gate on the host paths in a compose model
+# the sandbox hands the broker; handle_compose runs the real `docker compose`
+# as soon as it returns 0. Two things are under test here.
+#
+# Containment: a textual prefix test is not it. It lets /host/work-secret
+# through while the workdir is /host/work, and `docker compose config` hands
+# the validator paths it has NOT normalised (an absolute source keeps its `..`,
 # verified against compose 2.40.3) and never resolves symlinks. So the check
 # resolves each path and gates it with path_under(), and the cases below are
-# the three ways out: the sibling prefix, the `..` climb, and the symlink.
+# the four ways out: the sibling prefix, the `..` climb, the symlink, and the
+# DANGLING symlink that `test -e` calls missing.
+#
+# Coverage: bind sources are a fraction of the surface. A build context, an
+# absolute Dockerfile, an additional context, an ssh key, a local build cache,
+# a develop.watch path, a volume driver_opts.device, a secret or config file,
+# an env_file and a label_file each name a host path, and none of them passes
+# through .services[].volumes. The last two survive only the uninterpolated
+# render, and the default render is what reads them, so they are checked first
+# and the case below proves the default render never runs when one is outside.
 #
 # The real function is exercised against a stubbed `docker compose ... config`
 # so the resolved model is the input under test. The accept cases and the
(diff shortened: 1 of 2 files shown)


nickjalbert:
From review:

```
This is a substantial and well-argued hardening of the compose gate, and
the test file is the good kind of adversarial. The resolve-then-path_under
mechanic and the removal of workdir_alias check out (both sides resolved
before comparison, and the alias cases pin it). Five things before merge:

1. extra_contexts: emit *every* additional_contexts string value and let
   path_contained reject it, skipping only known non-path schemes
   (docker-image://, service:, oci-layout://, git/http(s)/ssh URLs). As
   written, select(... and startswith("/")) drops non-absolute values
   before any check, so additional_contexts fails OPEN while every other
   kind fails closed.
2. PATHS_PROGRAM/ENV_FILE_PROGRAM: extract or explicitly reject the
   remaining host-path surfaces — build.ssh (id=/path),
   build.cache_from/cache_to (type=local,src=/dest=), build.env_file,
   services[].label_file, develop.watch[].path, extends.file, top-level
   include. Same class as the fields added here: host paths that never
   pass through services[].volumes. Also narrow the "every host path the
   model names" comments to what is actually covered — an overclaim in a
   perimeter gate is worse than a gap.
3. compose_validate_model: run the --no-interpolate render and contain
   the env_file paths BEFORE the default render, and stop relaying
   compose's stderr verbatim to the caller (log it host-side, return a
   generic message). Today the default render reads the named host file
   first, and a dotenv parse error quotes the offending line back to an
   untrusted requester.
4. resolve_for_containment: stop the walk on symlinks
   (while [[ ! -e "$path" && ! -L "$path" && "$path" == */?* ]]) so a
   dangling symlink hits realpath, fails, and is rejected instead of
   being re-attached as a "missing" tail component.
5. tests: make the docker stub serve a different fixture when
   --no-interpolate is in argv, and add a case where the second render
   exits nonzero. Right now both renders return the same JSON, so the
   env_file cases would pass even with the second render deleted.
```

With those changes I'll think we'll be good to merge!  Thanks for the PR and let me know if any of that doesn't track.

beardthelion:
All five are in. Notes on what I found reproducing each against compose 2.40.3, since two of them turned out slightly different from the description.

**1. extra_contexts.** Changed as asked: every value now goes to `path_contained`, and only `docker-image://`, `oci-layout://`, `service:` and git/http(s)/ssh URLs are passed over. I could not turn the old filter into an actual escape though. Compose resolves a relative additional context against the project directory before the validator sees it (`rel: ../outside` renders as `/tmp/outside`), so nothing reached that `startswith("/")` test still relative. Worth doing anyway for the reason you gave: a gate that recognises what is safe fails open the moment the renderer changes.

**2. The remaining surfaces.** Four needed extraction and now have it: `build.ssh` `key=`, `cache_from`/`cache_to` `type=local,src=`/`dest=`, `develop.watch[].path`, and `label_file`. `label_file` behaves exactly like `env_file` (the default render reads it and inlines the labels, so the path survives only in the uninterpolated one), so it joins that program.

Two of the fields you listed did not need code, and the comment now says so rather than claiming full coverage:

- `build.env_file` is not in the 2.40.3 schema. Compose rejects the model: `services.s.build additional properties 'env_file' not allowed`.
- `extends.file` and top-level `include` are merged away before either render, so no path reaches the validator. What they pull in becomes ordinary services, and an `extends.file` naming `/etc` as a bind source is rejected as a bind. The residual is that compose reads the named YAML while rendering, which is what item 3 covers.

One thing to know about `cache_to`: compose leaves its `dest=` relative in the render (unlike every other path field), so a relative dest inside the workdir is rejected as non-absolute. Fail-closed, but it is a false rejection rather than a clean pass.

**3. Order and stderr.** Both done, and the leak is worse than a quoted line. Compose puts the file's contents in the error text:

```
env_file: [/etc/anything]  ->  failed to read /etc/anything: line 2: unterminated quoted value "<the value>"
                           ->  unexpected character in variable name "<the whole line>"
```

That is a read primitive for anything the broker user can point at. The uninterpolated render now runs first and its `env_file`/`label_file` paths are contained before the render that opens them, and a render failure is logged host-side with a fixed message going back. The cost: an error in the requester's own compose file now comes back as a generic rejection with rc 65 instead of compose's own text and exit code.

**4. resolve_for_containment.** Added the `! -L` test. Reproduced first: a symlink inside the workdir pointing at a non-existent target outside it was accepted, and so was any path through it.

**5. The stub.** Rebuilt. It answers the two renders from separate fixtures, can fail either one independently, and records which ran, which is what lets the new case assert that an env file outside the workdir loses *before* the default render happens. Both new failure cases are there.

Three of the cases drive a whole request through `handle_compose` rather than calling the validator directly, since a gate that rejects correctly while the response reads as empty or successful is still a hole. They pin what the sandbox actually receives (rc 65, the containment message, no stdout), that compose's own words are not in it, and that they *are* in the broker log, which is what makes swallowing them redaction rather than loss.

49 cases in all, and every guard is confirmed load-bearing: I removed each one in turn and watched the matching case go red (the dangling symlink, the relative additional context, each of the four new fields, the uninterpolated render itself, the stderr relay, the render order, the response the caller gets, and the log line). The extraction programs were also run against real `docker compose config` output rather than only handcrafted fixtures.

One limit worth stating: for the four fields added here I confirmed that compose renders a host path and that the broker now rejects it, but I did not build with a local cache pointing outside the workdir and watch buildkit write there. That is a weaker claim than the one behind the volume-device and secret cases in the earlier commits, which were observed escaping.

nickjalbert:
Thanks!  Merging!

PR #48, opened 2026-08-25
Fix dispatcher stalls caused by unreaped thinker zombies

wtfsayo:
## Summary

- distinguish Linux zombie step children from live thinker processes
- reap completed async children and prune the in-memory and on-disk step ledgers
- keep active-step counts in the dispatcher shell instead of losing cleanup state through command substitution
- add a regression test proving a scheduled wake fires after the previous child exits

## Root cause

The dispatcher used `kill -0` as its liveness test. An exited but unreaped Linux child remains in the process table with state `Z`, so `kill -0` succeeds and `_thinker_busy` suppresses all later wakes for that thinker. `_count_active_steps` also ran through command substitution at every call site, which discarded its `_step_entries` mutations in a subshell.

## Verification

- `bash -n bin/thinkers tests/test_thinkers_wake_at.sh`
- `bash tests/test_thinkers_wake_at.sh` on macOS: 14 passed, 0 failed
- same focused test in a fresh Linux container: 14 passed, 0 failed
- `bash tests/run-all.sh`: 21 suites passed, 0 failed
- `shellcheck -S warning bin/thinkers tests/test_thinkers_wake_at.sh`
- `git diff --check`

### DIFF
--- a/bin/thinkers
+++ b/bin/thinkers
@@ -467,27 +467,60 @@ _start_dispatcher() {
             _wd_last="${_new}${_new:+ }${_tname}:${SECONDS}"
         }
 
-        # Reap finished entries; print count of still-active ones
-        _count_active_steps() {
-            local _count=0 _entry
-            local _new=""
+        # Reap finished step children and prune both the in-memory and on-disk
+        # process lists. `kill -0` alone is insufficient on Linux: an exited,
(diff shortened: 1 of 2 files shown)


nickjalbert:
LGTM!  Thanks!

PR #49, opened 2026-08-25
llm: document and test env-set LLM_PROVIDER

samvelkoch:
# llm: document and test env-set LLM_PROVIDER

Rebased on main. The one-line fix this PR opened with is already there via #67,
so what remains is the part that wasn't: documentation and regression tests.

## What's in it

- `bin/llm --help` lists `LLM_PROVIDER` under Environment. It points at
  `--provider` for the list of provider names rather than repeating it, so the
  two can't drift as providers are added — they already had, my original text
  predated OpenCode.
- `design/model-resolution.md` names `LLM_PROVIDER` in the knobs table and, in
  the "no hard provider dependency" paragraph, says how a model name that
  implies no provider is reached. The page previously described provider
  selection as name-only.
- `tests/test_llm_provider_env.sh` — new, in the style of `test_llm_retry.sh`
  (stubbed `curl` recording the URL it was called with, no network).

## The tests

Six behaviours, including the mismatch warning main just gained:

- an alias no pattern matches is reachable through `LLM_PROVIDER` and the
  request goes to the configured endpoint;
- that case warns nothing — `detect_provider` cannot classify the name, so
  there is nothing for the environment to contradict;
- `LLM_PROVIDER=openai` with `claude-sonnet-4-5` is honored *and* reports
  `overrides detected provider 'anthropic'` on stderr;
- `--provider` overriding the environment warns about nothing: the flag is a
  deliberate choice, not a mismatch;
- an unset provider on an unclassifiable name still dies with
  `Cannot detect provider`;
- `claude-*` and `vendor/model` detection unchanged.

`tests/test_llm_provider_env.sh`: 10 passed, 0 failed. `test_llm_retry.sh`
stays green.

LOC gate: +4 code lines in `bin/` (the `--help` entry), comfortably inside
`LOC_LIMIT: 10000`.

## Context

Found while running Headlong against a self-hosted Qwen behind an internal
gateway, where the model is served under the alias `qwen-auto`. Two other
things stood in the way there; happy to open them separately if useful:

- `get_model_max_tokens` falls back to 4096 for an unrecognized name, which
  truncates utility calls that ask for more (the persona writer's `llm -t 3000`
  came back empty, so `headlong-init` silently used the starter template).
- vLLM-style endpoints take `chat_template_kwargs.enable_thinking` rather than
  `reasoning.effort`; with reasoning left on, that same budget goes to reasoning
  tokens.

### DIFF
--- a/bin/llm
+++ b/bin/llm
@@ -202,6 +202,10 @@ Environment:
                            opencode-go/* bills the Go subscription, opencode/*
                            bills Zen. Keys: `opencode auth login`
   LLM_API_URL              Override provider API URL
+  LLM_PROVIDER             Provider to use when the model name implies none
+                           (same names as --provider); --provider takes
+                           precedence, and a value that disagrees with the
+                           name warns on stderr
   LLM_RETRIES              Retries for transient API failures (default: 2;
                            0 disables). Never retries after partial output.
   LLM_RETRY_BACKOFF        Backoff base seconds between retries (default: 1)
--- a/design/model-resolution.md
+++ b/design/model-resolution.md
@@ -15,7 +15,10 @@ environment variables that feed that decision come from. Two layers:
 There is no hard provider dependency anywhere: `bin/llm` picks the
 provider from the model name (`claude-*` → Anthropic, `vendor/model` →
 OpenRouter, `gpt-*`/`o*` → OpenAI, `gemini-*` → Gemini) and each provider
-needs only its own `<PROVIDER>_API_KEY`. Hardcoded `claude-*` names below
+needs only its own `<PROVIDER>_API_KEY`. A model name that implies no
+provider — a local OpenAI-compatible alias, say — is reached by setting
+`LLM_PROVIDER` or passing `--provider`; an env value that contradicts a
+name the harness *can* classify is still honored, but says so on stderr. Hardcoded `claude-*` names below
 are last-resort defaults, reached only when nothing is configured.
 
 ## The knobs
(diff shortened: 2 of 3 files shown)


nickjalbert:
Hello!  

The change you proposed (honoring `LLM_PROVIDER` from the environment) just landed on main as part of #67, which included the same one-line fix. Your diagnosis was right and the behavior you wanted is now in.

If you could rebase your PR, I'd still like to take your text updates and tests. 

One heads up: `main` also just gained a warning that prints to stderr when an environment-set `LLM_PROVIDER` disagrees with what the model name implies.  One or two of your test cases may want to assert that warning rather than silence.

Thanks for the contribution!

samvelkoch:
Thanks — rebased on main, and I dropped the one-line fix since #67 has it. What's left is the documentation and the tests.

Good call on the warning; the tests now assert it in three places:

- an alias no pattern matches is reachable through `LLM_PROVIDER` and warns *nothing* — `detect_provider` can't classify the name, so there's nothing for the environment to contradict;
- `LLM_PROVIDER=openai` with `claude-sonnet-4-5` goes to OpenAI **and** reports `overrides detected provider 'anthropic'` on stderr;
- `--provider` overriding the environment warns about nothing, since the flag is a deliberate choice rather than a disagreement.

`tests/test_llm_provider_env.sh`: 10 passed, 0 failed.

One small change while I was in `--help`: the `LLM_PROVIDER` entry points at `--provider` for the list of provider names instead of repeating it. My original text had already drifted — it predated OpenCode.

`bin/` grows by 4 code lines (the `--help` entry), well inside `LOC_LIMIT`.

nickjalbert:
LGTM, thank you!

PR #51, opened 2026-08-25
Thinkers: read the API key from the state home the installer writes to

### LINKED ISSUE #50: Thinkers never read the API key from the renamed state home
`tools/headlong-init` writes the provider API key and `SHELLM_MODEL` to `$HEADLONG_HOME/.env` (`~/.headlong/.env`) and nowhere else. `thinkers/_lib/common.sh`'s `_require_env` fills its defaults from the identity `.env`, `./.env` and `$HOME/.shellm/.env`, which was the state home before the rename. The rename (99cb54d) moved `bin/llm` and `bin/shellm` onto `~/.headlong` and left this file behind.

The step scripts cannot fall through to `llm` for it, which is what makes this bite rather than being cosmetic. Both launchers export `SHELLM_HOME` as `<identity>/.shellm` (`bin/thinkers:156`, `web/src/headlong_web/control.py:69`), and `bin/llm:33` and `bin/shellm:31` resolve the state home as `${HEADLONG_HOME:-${SHELLM_HOME:-$HOME/.headlong}}`, so inside a thinker they read the identity directory rather than the file holding the key.

With the key only in `~/.headlong/.env`, which is what a fresh install produces:

```
OPENROUTER_API_KEY after _require_env: <unset>
SHELLM_MODEL       after _require_env: <unset>
forwarded to nested shellm: no
```

and `llm` run with the `SHELLM_HOME` a thinker gets:

```
llm: error: No model specified and no API key found. Set ANTHROPIC_API_KEY, OPENAI_API_KEY, GEMINI_API_KEY, or OPENROUTER_API_KEY, or pass -m MODEL.
```

Unset `SHELLM_HOME` and the same `llm` finds the key, which is the control for both.

So a thinker only works when something upstream has already exported the key. A provisioned box is fine: `deploy/thinkers-service.sh` sources `$APP_DIR/.env` before starting the dispatcher. A local install is not, unless the dash happens to have been started from a shell that had loaded `~/.headlong/.env`. Start it fresh (after a reboot, or `headlong-web` from a plain shell) and the thinkers run with no key.

Worth noting `tools/headlong-init:139-140` describes its own env layering as "same layering as bin/llm and the thinkers", which is what the thinkers should be doing and currently are not.

A fix is on the way: add the framework state home to `_require_env`'s lookup, resolved without `SHELLM_HOME` for the reason above, with `~/.shellm` kept after it for a pre-rename install.

beardthelion:
Fixes #50.

`tools/headlong-init` writes the provider key and `SHELLM_MODEL` to `$HEADLONG_HOME/.env` and nowhere else. `_require_env` filled its defaults from the identity `.env`, `./.env` and `$HOME/.shellm/.env`, the state home from before the rename; 99cb54d moved `bin/llm` and `bin/shellm` onto `~/.headlong` and left this file behind.

The step scripts cannot lean on `llm` to load it instead, which is what turns this from cosmetic into a thinker that does not run. Both launchers export `SHELLM_HOME` as `<identity>/.shellm` (`bin/thinkers:156`, `control.py:69`), and `llm` and `shellm` resolve the state home through `SHELLM_HOME`, so inside a thinker they read the identity directory. Checked by running it: with the key only in `~/.headlong/.env`, a step resolved no key and no model and forwarded neither to the nested `shellm`, and `llm` under that same `SHELLM_HOME` died with `No model specified and no API key found`. Unset `SHELLM_HOME` and the same `llm` finds it, which is the control.

So the lookup here is resolved as `${HEADLONG_HOME:-$HOME/.headlong}/.env`, deliberately without `SHELLM_HOME` unlike the two tools, with `~/.shellm` kept after it for a pre-rename install. Values still fill in only what is unset, so the dispatcher's environment keeps winning. That is why a provisioned box never showed this: `deploy/thinkers-service.sh` sources `$APP_DIR/.env` before starting. A local install has nothing playing that role, so a dash started from a shell that never loaded `~/.headlong/.env` gives every thinker an empty key.

## Verification

`tests/test_thinker_env_fallback.sh` is new; nothing in `tests/` sourced `common.sh` before, which is why this survived the rename. Eight cases, four of which fail without the change:

- the state home `.env` is read, and beats the legacy path
- `HEADLONG_HOME` redirects the lookup
- the key reaches the nested `shellm` as a bare `--var NAME`

The other four pass before and after, and are there so the file cannot go green vacuously: the legacy `~/.shellm/.env` still works, the identity `.env` still wins over both, no env file leaves the key unset, and an already-set value still beats the file.

`tests/run-all.sh` is 22/22 and `shellcheck -S warning` is clean on both files. No bash 4 constructs, so the macOS job is covered.

### DIFF
--- a/tests/test_thinker_env_fallback.sh
+++ b/tests/test_thinker_env_fallback.sh
@@ -1,0 +1,125 @@
+#!/usr/bin/env bash
+# tests/test_thinker_env_fallback.sh — a thinker finds the API key where the
+# installer actually wrote it.
+#
+# tools/headlong-init writes the provider key and SHELLM_MODEL to
+# $HEADLONG_HOME/.env and nowhere else. The thinker step scripts resolve those
+# BEFORE calling llm/shellm (which load .env too late to influence the -m
+# flag), so _require_env has to read that file. It cannot lean on llm to do it
+# instead: both launchers export SHELLM_HOME as <identity>/.shellm
+# (bin/thinkers, web control.py), and llm resolves the state home through
+# SHELLM_HOME, so under a thinker it reads the identity directory rather than
+# the framework state home.
+#
+# The legacy ~/.shellm case is here because a pre-rename install keeps its
+# state there, and the "already set wins" case because the dispatcher's own
+# environment (deploy/thinkers-service.sh, the web _ENV_WRAPPER) must always
+# beat a file.
+
+set -uo pipefail
+REPO="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+
+pass=0
+fail=0
+ok()  { pass=$((pass+1)); printf 'ok   %s\n' "$1"; }
+bad() { fail=$((fail+1)); printf 'FAIL %s%s\n' "$1" "${2:+ — $2}"; }
+
+# probe <name> <expected-key-value> [env-assignment ...] -- runs _require_env in
+# a fresh HOME laid out by the caller's setup function.
+probe() {
+    local name="$1" want="$2" setup="$3" ; shift 3
+    local out
+    out=$(
+        H=$(mktemp -d)
+        export HOME="$H"
+        mkdir -p "$H/id/memories" "$H/id/skills" "$H/id/kernel" "$H/id/trajectories" "$H/wd"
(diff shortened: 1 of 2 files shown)


nickjalbert:
LGTM, thanks!

PR #53, opened 2026-08-25
install: keep API keys off the docker command line

### LINKED ISSUE #52: Container install passes API keys on the docker command line
The container install path forwards the operator's environment into the container by appending `-e VAR=value` to `docker run` (`install.sh:161-165`, used at 172-175). That includes `ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, `GEMINI_API_KEY` and `OPENROUTER_API_KEY`, so the key is an argument of the docker client and readable from the process table by any other user on the machine for as long as that client runs:

```
$ ps -axo args= | grep sk-or-
docker run --rm -e OPENROUTER_API_KEY=sk-or-... -e HEADLONG_IDENTITY_NAME=ada ...
```

This is the rule the repo already holds itself to everywhere else. `tests/test_var_secrets.sh` exists to assert that "the secret's value appears in no process's argv (`ps`)" for `shellm --var`, and `bin/thinkers:267-272` forwards keys to nested calls by bare name with the comment "Keys go by NAME (bare `--var NAME`): shellm reads the value from its environment, so it never shows up in `ps`". The installer is the one place that passes values instead.

Worth flagging for anyone fixing it: switching the whole loop to name-only `-e VAR` is not safe. Docker reads a name-only `-e` from the client's own environment, so the variable has to be exported, and `install.sh:30-31` assigns `HEADLONG_REPO` and `HEADLONG_BRANCH` without exporting them. Forwarding those by name sends nothing, and the in-container installer then clones the default repo and branch instead of the operator's. Checked both ways:

```
set but NOT exported, -e K  ->  inside=[]
exported,             -e K  ->  inside=[sk-or-EXPORTED]
```

So the keys want name-only plus an explicit export, and the non-secret interview answers want to keep their values inline.

A fix is on the way.

beardthelion:
Fixes #52.

The container install forwarded everything it carries as `-e VAR=value`, so an API key spent the life of the docker client readable in `ps` by any other user on the machine. `bin/thinkers:267-272` already forwards keys to nested calls by bare name for exactly this reason, and `tests/test_var_secrets.sh` pins the rule for the shellm path; the installer was the one place doing it the other way.

Keys now go by name with an explicit export, and the interview answers keep their values inline. That split is not cosmetic: docker reads a name-only `-e VAR` from the client's own environment, so the name has to be exported, and `install.sh:30-31` assigns `HEADLONG_REPO` and `HEADLONG_BRANCH` without exporting them. Forwarding those by name would send nothing and the in-container installer would clone the default repo instead of the operator's fork. They are not secrets, so inline is the right place for them.

The arguments move into `_docker_forward_args` because `_offer_docker_install` reads from `/dev/tty` and cannot be driven from a test.

## Verification

End to end against a real daemon, with the arguments the helper produces:

```
after:  inside container: key=[sk-or-E2ECANARY] name=[ada] repo=[https://example.invalid/fork.git]
        ps exposure:      none
before: inside container: key=[sk-or-E2ECANARY]
        ps exposure:      YES -> docker run --rm -e OPENROUTER_API_KEY=sk-or-E2ECANARY ...
```

The unexported repo override in that first line is the case the simpler fix would have dropped.

`tests/test_install_secret_argv.sh` is new: no key value on the command line, the key forwarded by name, the key actually exported so the bare name carries, answers still inline, an unexported repo and branch override still reaching the container, and no key forwarded when none is set. Three of the seven fail against the old behaviour.

`tests/run-all.sh` is 22/22, `tests/smoke_install.sh` is 114/114, and `shellcheck -S warning` is clean. No bash 4 constructs.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -96,6 +96,33 @@ _docker_daemon_ok() {
     command -v docker >/dev/null 2>&1 && docker info >/dev/null 2>&1
 }
 
+# The `docker run` arguments that carry the operator's environment into the
+# container, one per line for the caller to read into an array.
+#
+# Keys go by NAME. `-e VAR=value` puts the value in the docker client's argv,
+# where `ps` shows it to every other user on the machine for as long as the
+# client runs; bin/thinkers forwards keys the same way for the same reason, and
+# tests/test_var_secrets.sh pins the rule for shellm. Docker reads a name-only
+# `-e VAR` from its own environment, so the name must be exported, which is why
+# the answers below do NOT use it: install.sh assigns HEADLONG_REPO and
+# HEADLONG_BRANCH itself without exporting them, so a bare name would forward
+# nothing and the container would clone the default repo instead of the
+# operator's. They are not secrets, so inline is fine.
+_docker_forward_args() {
+    local var
+    for var in ANTHROPIC_API_KEY OPENAI_API_KEY GEMINI_API_KEY OPENROUTER_API_KEY \
+               OPENCODE_API_KEY; do
+        if [[ -n "${!var:-}" ]]; then
+            export "${var?}"
+            printf '%s\n%s\n' "-e" "$var"
+        fi
+    done
+    for var in HEADLONG_IDENTITY_NAME HEADLONG_IDENTITY_VIBE HEADLONG_IDENTITY_FOCUS \
+               HEADLONG_IDENTITY_USER HEADLONG_OPERATOR_NAME HEADLONG_REPO HEADLONG_BRANCH; do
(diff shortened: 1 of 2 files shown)


nickjalbert:
#67 just landed and touched the same key-forwarding list in install.sh (it added OPENCODE_API_KEY), so this branch needs a rebase before it can merge.  looking forward to it!

beardthelion:
Rebased onto main, and `OPENCODE_API_KEY` is in the forwarded list now. Green and mergeable.

nickjalbert:
LGTM. Thanks!

PR #55, opened 2026-08-25
responder: answer "already handled?" from the trajectory tail

### LINKED ISSUE #54: Responder scans the whole trajectory before every reply
`thinkers/responder/step:81` runs its idempotency check with `traj cat "${ROOT_TRAJ_ID:-$TRAJ_ID}" --raw` over the whole root trajectory, before any reply is composed. `traj cat` emits every line, so the cheapest thing the mind does costs O(total lifetime log size), on the path someone is waiting on.

Measured with the step's own jq program over real `traj cat` output, one machine, growing the same trajectory:

| steps | size | full `traj cat` | `tail -n 5000` |
|---|---|---|---|
| 20k | 13 MB | 333 ms | 81 ms |
| 60k | 40 MB | 979 ms | 84 ms |
| 120k | 80 MB | 1,948 ms | 81 ms |
| 240k | 161 MB | 3,928 ms | 83 ms |
| 460k | 308 MB | 7,398 ms | 80 ms |

`design/monolith_run_health.md` records a live trajectory at 312 MB, so this is roughly 7 seconds of dead time per inbound message on an identity that has been running a while.

The bounded reader already exists for exactly this: `_root_traj_raw_tail` in `thinkers/_lib/common.sh:143`, whose comment names the same cost ("traj cat is O(file size) ... 78s context builds on a 532MB file"). `fe2acd2` moved the monolith's work probe onto it, naming "two full 312MB scans per wake on cleo". `monolith/step` and `_recent_stream` both use it. The responder is the last `traj cat --raw` left in `thinkers/`.

One thing worth flagging for whoever fixes it: a plain tail is not a drop-in. The check's fourth clause locates the trigger step by index (`$in.key + 1`) to find later messages to the same person, and the first three clauses are keyed on the trigger too. A window is sound only because every record it looks for can only be appended after the trigger, so a window holding the trigger holds the whole answer. A window that misses the trigger, which is what a redelivered old message looks like, would report an answered message as unanswered and the identity would reply twice. That case needs the full scan.

A fix is on the way.

beardthelion:
Fixes #54.

The idempotency check ran over the whole root trajectory before any reply was composed, so the cheapest thing the mind does cost O(total lifetime log size), on the path someone is waiting on. Measured with the step's own jq program over real `traj cat` output: 0.98s at 40MB, 3.9s at 161MB, 7.4s at 308MB, against ~0.08s for the tail. `design/monolith_run_health.md` records a live trajectory at 312MB.

`_root_traj_raw_tail` already existed for this, and `fe2acd2` moved the monolith's work probe onto it naming "two full 312MB scans per wake on cleo". The responder was the last `traj cat --raw` in `thinkers/`.

A window is sound here because every record the check looks for (a stamped reply, a decision observation, a live claim, a later message to the same person) can only be appended after the trigger step, so a window holding the trigger holds the whole answer. The case that is not sound is a window that misses the trigger, which is what a redelivered old message looks like: the check would report an answered message as unanswered and the identity would reply twice. So the scan reports "trigger not in this stream" separately from a verdict, and only that runs the full `traj cat`. At 308MB: 96ms when the trigger is in the window, 8.6s and correct when it is not.

The scan moves into `_responder_already_handled` beside `_root_traj_raw_tail`. That is what makes it testable at all: the step script runs top to bottom and cannot be sourced, which is why none of this had a test.

## Verification

`tests/test_responder_idempotency.sh` is new. Six cases pin the verdicts that must not move: a stamped reply, a decision observation and a fresh claim all count as handled; a stale claim deliberately does not (that is the crashed send, and treating it as handled is how messages used to drop); a later unstamped message to the same person counts; an earlier one does not. Two more cover the change: no `traj cat` at all when the trigger is in the window, and a full scan that still finds an old reply when it is not. A `traj` shim records the subcommands, so "did it read the whole file" is answered by what was invoked rather than by timing.

I extracted the helper with the original full-scan behaviour first and confirmed all six verdicts green against it, so the window change is a real red-then-green rather than a test written to fit the new code.

`tests/run-all.sh` is 22/22, `shellcheck -S warning` is clean, no bash 4 constructs.

Note this touches `thinkers/_lib/common.sh`, as does #51. Different parts of the file; I merged the two branches locally and they combine with no conflict, 23/23 with both new tests green.

### DIFF
--- a/tests/test_responder_idempotency.sh
+++ b/tests/test_responder_idempotency.sh
@@ -1,0 +1,151 @@
+#!/usr/bin/env bash
+# tests/test_responder_idempotency.sh — the responder's "already handled?"
+# check, and the window it reads to answer it.
+#
+# Every inbound message runs this check before a reply is composed, so it sits
+# on the latency path. It used to scan the whole root trajectory: measured at
+# 0.98s over 40MB and 7.4s over 308MB, against ~0.08s for the tail, and
+# design/monolith_run_health.md records real trajectories in that range.
+#
+# Reading a window is only sound because every record the check looks for (a
+# stamped reply, a decision observation, a live claim, a later message to the
+# same person) can only be appended AFTER the trigger step. So a window holding
+# the trigger holds everything relevant to it; a window that does NOT is the
+# case that must fall back to the full scan, or a redelivered old message gets
+# answered twice.
+#
+# The six verdict cases are pinned first because they are the behaviour the
+# bounded read must not change.
+
+set -uo pipefail
+REPO="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+export PATH="$REPO/bin:$PATH"
+
+pass=0
+fail=0
+ok()  { pass=$((pass+1)); printf 'ok   %s\n' "$1"; }
+bad() { fail=$((fail+1)); printf 'FAIL %s%s\n' "$1" "${2:+ — $2}"; }
+
+command -v jq >/dev/null 2>&1 || { echo "FAIL jq not found — the check needs it, so this proves nothing"; exit 1; }
+
+WORK=$(mktemp -d)
+trap 'rm -rf "$WORK"' EXIT
+
+ME=ada
+THEM=nick
+TRIGGER=trig-1
+FRESH=$(date -u +%Y-%m-%dT%H:%M:%S)
+STALE=2000-01-01T00:00:00
+
+# step <type> <json-fields...> — one trajectory line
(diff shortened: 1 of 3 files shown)


nickjalbert:
From review:

```
This is a well-argued change, and the header comment explaining 
*why* a window that holds the trigger holds the whole answer is 
exactly the kind of note that keeps this code safe to touch later. 
The tail-first / fallback split is sound and the six pinned verdicts 
plus both I/O paths are good coverage. 
```

Merging as is; I'll handle a few minor follow-ups shortly.  Thanks!!!

PR #57, opened 2026-08-25
web: bound the dispatch endpoint's read, not just its answer

### LINKED ISSUE #56: Dispatch endpoint reads the whole dispatcher.log to return its last 2000 events
`parse_dispatch_log` (`web/src/headlong_web/logs.py:44`) takes `max_events=2000` and returns `events[-max_events:]`, but gets there with `log_path.read_text(...).splitlines()` and a regex pass over every line. The cap bounds the answer, not the work.

Measured with the real function against a synthetic dispatcher.log of realistic lines:

| lines | size | parse | peak allocation | events returned |
|---|---|---|---|---|
| 50k | 3.2 MB | 485 ms | 17 MB | 2000 |
| 200k | 12.7 MB | 1,977 ms | 68 MB | 2000 |
| 800k | 50.6 MB | 8,219 ms | 271 MB | 2000 |
| 3M | 190 MB | 31,348 ms | 1,013 MB | 2000 |

`deploy/MIGRATIONS.md:98` records 465M of dispatcher logs on a real box, and nothing rotates them.

The cost is not paid once per look. `web/viewer/app/routes/thinkers.tsx:485` polls `GET /api/identities/{id}/dispatch` on `refetchInterval`, and `web/viewer/app/lib/api.ts:441` sets that to 2000 ms. On a grown identity a viewer sitting on the thinkers tab starts a fresh multi-second, multi-hundred-MB request every two seconds, and they overlap in the threadpool. Nobody has to be hostile for that to happen; it is the normal way the page is used. Worth noting `--read-only`, whose help text presents it as the hardening for binding beyond 127.0.0.1, does not take this endpoint away, because it is a read.

`tail_log` twenty lines above in the same file already does the bounded thing for raw log tails, seeking to `total - tail_bytes` and dropping the partial first line. `parse_dispatch_log` just never got the same treatment.

A fix is on the way: read backward until `max_events` lines are in hand, then parse those. Same shape, same ordering, same cap; 26 ms and 0.9 MB on the 190 MB log.

beardthelion:
Fixes #56.

`parse_dispatch_log` advertised `max_events=2000` and returned `events[-2000:]`, but read the whole file and ran a regex pass over every line to get there. The cap governed the answer and nothing else. Measured with the real function: 2.0s and 68MB at 12.7MB, 8.2s and 271MB at 50MB, 31.3s and 1013MB at 190MB, each returning the same 2000 events. `deploy/MIGRATIONS.md` records 465M of dispatcher logs on a real box, and nothing rotates them.

That is not paid once per look. `thinkers.tsx` polls this endpoint on `refetchInterval` and `api.ts` sets it to 2000ms, so on a grown identity a viewer sitting on the tab starts a fresh multi-second, multi-hundred-MB request every two seconds and they overlap in the threadpool. `--read-only` does not take the endpoint away either, since it is a read.

`tail_log` twenty lines above already seeks to the end for raw log tails. This does the same, counting lines rather than bytes because the cap is in events: read backward in 64KB chunks until `max_events` non-blank lines are in hand, drop the partial first line, parse those. Same shape, same ordering, same cap. On the 190MB log: 26ms and 0.9MB, still 2000 events.

## Verification

`web/tests/test_dispatch_log.py` is new; `logs.py` had no tests at all, which is part of why a cap could read like a bound. Nine cases: each event kind parses, newest-last ordering and the cap, blank lines, a missing trailing newline, an empty file, a missing file, a multibyte character split across a chunk boundary, and a line longer than the chunk. Two of them are about the change itself, and both fail on the old code: `Path.read_text` is monkeypatched to raise so a whole-file read fails outright, and peak allocation has to stay under a quarter of the file size.

The seven behavioural cases were written against the old implementation and passed there, so they pin existing semantics rather than describing the new code. On top of that, a differential run against the original: 160 comparisons over random logs (0 to 20k lines, blank lines, missing trailing newlines, caps of 1/7/100/2000), zero mismatches.

`web/tests` is 159 passed / 2 skipped.

### DIFF
--- a/web/src/headlong_web/logs.py
+++ b/web/src/headlong_web/logs.py
@@ -41,13 +41,46 @@ def tail_log(log_path: Path, tail_bytes: int) -> dict[str, Any]:
     }
 
 
+def _tail_lines(log_path: Path, max_lines: int, chunk_bytes: int = 1 << 16) -> list[str]:
+    """The newest max_lines non-blank lines, read backward from the end.
+
+    dispatcher.log is never rotated and the endpoint above this is polled every
+    couple of seconds, so the whole-file read this replaces cost 31s and ~1GB of
+    allocation on a 190MB log to return the same 2000 events. tail_log solves
+    the same problem for raw log tails; this one counts lines rather than bytes,
+    because the caller's cap is in events.
+    """
+    with log_path.open("rb") as fh:
+        fh.seek(0, 2)
+        pos = fh.tell()
+        data = b""
+        while pos > 0:
+            step = min(chunk_bytes, pos)
+            pos -= step
+            fh.seek(pos)
+            data = fh.read(step) + data
+            # Only whole lines count, so ignore anything before the first
+            # newline while more of the file remains.
+            complete = data.split(b"\n", 1)[-1] if pos > 0 else data
+            if sum(1 for line in complete.splitlines() if line.strip()) >= max_lines:
+                break
+    text = data.decode("utf-8", errors="replace")
+    lines = text.splitlines()
+    if pos > 0 and lines:
+        lines = lines[1:]  # drop the (likely partial) first line
+    return [line for line in lines if line.strip()][-max_lines:]
+
+
 def parse_dispatch_log(identity_dir: Path, max_events: int = 2000) -> list[dict[str, Any]]:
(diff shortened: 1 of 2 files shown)


nickjalbert:
Thanks!!

PR #69, opened 2026-08-26
init/llm: resolve the provider from the key, not the variable holding it

korstoilija:
Two small, independent bugs that both show up as a non-Anthropic provider failing with a message that points somewhere other than the actual cause. Found while installing with an OpenRouter key.

## 1. A key in the wrong variable silently picks the wrong provider

`_have_key()` returns true if *any* `*_API_KEY` is set, and `_first_key_var()` then takes the first one in a fixed order. Nothing checks that the key actually belongs to the variable holding it.

Repro:

```bash
export OPENAI_API_KEY=sk-or-v1-...   # an OpenRouter key
curl -fsSL https://headlong.ai/install.sh | bash
```

`_first_key_var` returns `OPENAI_API_KEY`, `_default_model_for` gives `gpt-5.5`, `detect_provider` maps `gpt-*` to `openai`, and the handshake goes to `api.openai.com`:

```
==> Checking the LLM connection (model: gpt-5.5)
    FAILED — OPENAI_API_KEY (sk-o...bd64) did not get a reply from gpt-5.5.
    The key may be mistyped, for another provider, or out of credit.
```

The key is fine. It is in the wrong variable, and nothing in the output says so — the masked hint even elides the `sk-or-` prefix that would give it away. `gpt-5.5` never appears in anything the user typed, which makes it hard to work back to what happened.

`_prompt_key` already derives the provider from the prefix via `_key_var_for`, and the comment above `_env_set "$HEADLONG_HOME/.env" "SHELLM_MODEL" ...` says the intent is that "a stale key for another provider in the env can't shadow the one we just validated". This extends the same reasoning to keys arriving from the environment rather than from a paste.

Behaviour now:

- `_first_key_var` prefers a variable its own key's prefix does not contradict, so a correctly placed key always beats a misplaced one.
- `_reconcile_key_vars` moves a misplaced key to where it belongs and says so on stderr.
- A key is only moved when its prefix is **unambiguous** — `sk-ant-`, `sk-or-`, `AIza`. A bare `sk-` key is OpenAI *or* OpenCode (`_prompt_key` has to ask), so it is treated as no evidence at all and never moved. That is what `_key_var_unambiguous` is for, and why it is deliberately stricter than `_key_var_for`.
- `_reconcile_key_vars` is called once from `_ensure_api_key` in the parent shell, never inside `$(...)`, since the `export` has to outlive the call.

If nothing is misplaced, nothing changes.

## 2. `LLM_MAX_TOKENS` was ignored

`bin/llm` had:

```bash
LLM_MODEL="${LLM_MODEL:-}"
LLM_SYSTEM=""
LLM_MAX_TOKENS=""          # <-- env and .env silently dropped
LLM_MESSAGES=""
LLM_STREAM=1
LLM_PROVIDER="${LLM_PROVIDER:-}"
```

so the variable was ignored from the environment *and* from the state `.env` that `_llm_load_env` sources, unlike its neighbours. Only `-t` worked.

This matters most where `get_model_max_tokens` falls through to the generic `*/*` cap of 16384. The comments on the `anthropic/*` and `x-ai/grok*` entries already explain why that starves a reasoning model — "reasoning tokens count against max_tokens, so 16384 starves any answer that needed heavy thinking first" — and any other vendor-prefixed reasoning model on OpenRouter hits exactly that, with no way to raise it short of editing `bin/llm`. One-line fix; I left the cap table alone.

## Testing

New `tests/test_key_var_resolution.sh`, in the style of `test_sandbox_gate.sh` — `curl`, `docker` and `llm` stubbed, no network.

- 10/10 pass with these changes.
- 6/10 fail without them. The 4 that pass either way are the guards against changing behaviour that already worked: a correctly placed OpenAI key keeps `gpt-5.5` and warns about nothing, and a bare `sk-` key in `OPENCODE_API_KEY` is left alone.
- `tests/run-all.sh`: 23 passed, 0 failed.
- `shellcheck -S warning` clean on all three files.

Key fixtures are shaped so they cannot match a real provider pattern — the first version of this branch was blocked by GitHub push protection for containing a convincing fake `sk-or-v1-`, which seemed worth avoiding in a repo where people will copy the test.

Happy to split this into two PRs, or drop either half, if you would rather take them separately.

### DIFF
--- a/bin/llm
+++ b/bin/llm
@@ -35,7 +35,7 @@ _llm_load_env "$_llm_home/.env" || true
 
 LLM_MODEL="${LLM_MODEL:-}"
 LLM_SYSTEM=""
-LLM_MAX_TOKENS=""
+LLM_MAX_TOKENS="${LLM_MAX_TOKENS:-}"
 LLM_MESSAGES=""
 LLM_STREAM=1
 LLM_PROVIDER="${LLM_PROVIDER:-}"
--- a/tests/test_key_var_resolution.sh
+++ b/tests/test_key_var_resolution.sh
@@ -1,0 +1,117 @@
+#!/usr/bin/env bash
+# tests/test_key_var_resolution.sh — provider selection when a key sits in the
+# wrong variable, and LLM_MAX_TOKENS coming from the environment.
+#
+# Usage: tests/test_key_var_resolution.sh
+#
+# Why: the provider is chosen from whichever *_API_KEY variable is set first,
+# and the model default follows from that. A key exported into the wrong
+# variable (an sk-or- OpenRouter key in OPENAI_API_KEY) therefore picked
+# gpt-5.5, which detect_provider routes to api.openai.com, and the run died
+# with a confusing "Incorrect API key provided". The prefix now wins, except
+# where it is genuinely ambiguous (a bare sk- key is OpenAI or OpenCode).
+# No network and no real daemon are touched; curl, docker and llm are stubs.
+# Key fixtures below carry only the prefix that matters and are deliberately
+# shaped so they cannot match a real provider pattern (GitHub push protection
+# rejects fixtures that look like the genuine article).
+
+set -uo pipefail
+HERE="$(cd "$(dirname "$0")" && pwd)"
+REPO="$(dirname "$HERE")"
+
+WORK=$(mktemp -d)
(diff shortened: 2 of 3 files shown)


nickjalbert:
LGTM! Thank you!

PR #72, opened 2026-08-26
sandbox: confine symlink mounts to the checkout and an opt-in allowlist

korstoilija:
## What

`symlink_mounts()` walks the workdir for symlinks and mounts the **parent directory** of each target into the next container. The agent has write access to that workdir, so the set of host directories the sandbox can see is effectively chosen by the agent.

```bash
ln -s ~/.ssh/id_ed25519 x     # next run mounts ~/.ssh
ln -s ~/.headlong/.env  x     # next run mounts the state home, where
                              # headlong-init writes the API key
```

I confirmed the first one against a live install: a broker-launched container could list `~/.ssh` and read `id_ed25519`.

## What it is, and isn't

Nothing breaks out of the container. There's no kernel exploit, no privilege escalation, no docker socket. The broker runs on the host and constructs the `-v` flag itself from agent-controlled input — a confused deputy, not an escape. Impact is host file **reads** only: no writes, no host code execution.

That said, the installer's own framing is "the agent's shell commands still run sandboxed in Docker", and for confidentiality that doesn't hold — the whole of `$HOME` is one `ln -s` away. It's also cheap to trigger, and the reachability isn't really "would the agent choose to do this": these agents read untrusted text, so a string in a fetched page or a checked-out README is enough to aim it, and the result goes wherever the configured model goes.

## Approach

Dropping the feature isn't an option — every thinker, skill and kernel entry reaches the identity through a symlink into the app checkout, so a blanket deny breaks the agent. Targets are confined to:

- **the app checkout** (the directory holding `.identities`), where all of Headlong's own symlinks point; and
- **directories the operator opts into** via `HEADLONG_SANDBOX_RO_DIRS`, a colon-separated list read from the environment or the state `.env`.

Opted-in directories are additionally mounted `:ro` into every run, so an operator can hand the agent project trees to read without a symlink dance, and without that grant widening to anything else. Refusals are logged.

Default behaviour is unchanged for anyone not setting the variable, except that agent-planted symlinks out of the checkout stop being followed.

```bash
# in ~/.headlong/.env
HEADLONG_SANDBOX_RO_DIRS=/Users/me/git_repos:/Users/me/notes
```

## Two details worth a reviewer's eye

- `allow_root` and `workdir` are canonicalized before comparison. Targets arrive from `readlink -f`, so without this a symlink anywhere in the install path — `/var` → `/private/var` on macOS — makes every legitimate target look external and refuses all of them. I hit exactly this: my first version passed the security assertions and silently broke the agent.
- `sandbox_ro_dirs` uses `read -r d || [[ -n "$d" ]]`. The split list has no trailing newline, so a single-entry list is otherwise dropped in full.

## Testing

`tests/test_sandbox_symlink_mounts.sh` exercises the functions directly — no docker daemon, no network, so it runs anywhere the rest of the bash suite does.

- 9/9 with this change.
- 6 of the 9 fail without it, and the failure output shows the secrets directory being mounted as `:ro`.
- `shellcheck -S warning` clean.
- `tests/run-all.sh`: all green.

One note in case you see it locally: `test_llm_retry.sh` fails for me only when my real `~/.headlong/.env` is visible, because `bin/llm` sources the state `.env` and picks up my configured `SHELLM_MODEL`. With `HEADLONG_HOME` pointed at a scratch dir it's 70/70, on this branch and on pristine `main` alike, so it's unrelated to this change — but the bash suite isn't quite as hermetic as the CI comment suggests for anyone with a real install on the same machine.

## Disclosure

Happy to have raised this in public rather than privately — it's read-only, needs local agent action, and the fix is in the same message. If you'd rather handle this class of thing through a security advisory instead, say so and I'll use that route next time.

### DIFF
--- a/tests/test_sandbox_symlink_mounts.sh
+++ b/tests/test_sandbox_symlink_mounts.sh
@@ -1,0 +1,100 @@
+#!/usr/bin/env bash
+# tests/test_sandbox_symlink_mounts.sh — which host paths the broker will mount
+# into the sandbox.
+#
+# Usage: tests/test_sandbox_symlink_mounts.sh
+#
+# Why: symlink_mounts() mounts the *parent directory* of every symlink target
+# it finds in the workdir, and the agent has write access to that workdir. So
+# `ln -s ~/.ssh/id_ed25519 x` used to be enough to get ~/.ssh mounted into the
+# next container, and `ln -s ~/.headlong/.env x` the state home holding the API
+# key — the sandbox handing over host files rather than being broken out of.
+# Targets are now confined to the app checkout plus whatever the operator opts
+# into via HEADLONG_SANDBOX_RO_DIRS.
+#
+# The functions are exercised directly; no docker daemon and no network.
+
+set -uo pipefail
+HERE="$(cd "$(dirname "$0")" && pwd)"
+REPO="$(dirname "$HERE")"
+BROKER="$REPO/tools/shellm-docker-broker"
+
+WORK=$(mktemp -d)
+trap 'cd /; rm -rf "$WORK"' EXIT
+pass=0; fail=0
+ok()  { pass=$((pass+1)); printf 'ok   %s\n' "$1"; }
+bad() { fail=$((fail+1)); printf 'FAIL %s%s\n' "$1" "${2:+ — $2}"; }
+
+# Pull just the functions under test out of the broker, which otherwise starts
+# serving when run.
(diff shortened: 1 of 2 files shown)


nickjalbert:
LGTM!  Good find, and thank you for the contribution! As the project matures we'll need a better process for handling reports like this, but for now a PR like this is great. 

Merging, with a small maintainer patch to follow.

PR #73, opened 2026-08-26
persona: resume starts the responder, not just the monolith

korstoilija:
## The bug

`headlong-init:663` starts a mind with:

```bash
thinkers start monolith responder
```

`cmd_start` in `tools/persona:206`, which is what `<name> start` runs after a `<name> stop`, started:

```bash
thinkers start monolith
```

So the first pause/resume in an identity's life drops the responder permanently. Nothing puts it back until the next `headlong-init`.

## Why it's worth a test rather than just a one-line fix

It fails silently, and expensively.

After the resume, the dispatcher starts, `status.json` reports mind `ok`, the dash reports `ok`, and the agent is visibly thinking — the monolith is running and its trajectory keeps advancing every few seconds. Everything you'd check looks healthy. `run/active_thinkers` contains just `monolith`, but you'd have to already suspect the problem to go and read it.

The only symptom is that messages from the human stop being handled by the thinker built to handle them promptly, and wait on the monolith's reasoning loop instead. On a reasoning model that's the difference between a few seconds and minutes. And because the `<name> stop` that caused it may have been hours earlier, the effect is completely detached from the cause — it reads as "the agent is stuck", or as an argument that you need to run several agents so one is always free to talk to you.

I hit it exactly that way: paused and resumed a fresh install, then spent a while working out why replies had gone from ~5s to minutes. After the fix, a round trip measured 5s again.

## The test

`tests/test_thinker_set_coupling.sh` derives the thinker set from both starters and asserts they agree. This is deliberately in the spirit of `test_unit_name_couplings.sh` — the failure mode has the same shape, a string in one place that has to keep matching a string in another, with nothing to complain when it stops. Comments are stripped before matching so prose mentioning the command can't satisfy it.

It also:
- names the responder explicitly, so removing it from *both* starters still fails rather than silently agreeing on the wrong set;
- checks every named thinker actually ships a `step` script, since a starter naming a thinker that isn't there is another silent no-op.

5/5 with this change, 2 of the 5 fail without it.

`shellcheck -S warning` clean. `tests/run-all.sh` green.

(As in #72: `test_llm_retry.sh` fails locally when a real `~/.headlong/.env` is visible, because `bin/llm` sources the state `.env` and falls back to `$HOME/.headlong`. With `HOME` and `HEADLONG_HOME` pointed at a scratch dir it's 70/70 on this branch and on pristine `main`. This branch touches only `tools/persona`, nothing llm-related.)

### DIFF
--- a/tests/test_thinker_set_coupling.sh
+++ b/tests/test_thinker_set_coupling.sh
@@ -1,0 +1,72 @@
+#!/usr/bin/env bash
+# tests/test_thinker_set_coupling.sh — headlong-init and persona must start the
+# same thinkers.
+#
+# Usage: tests/test_thinker_set_coupling.sh
+#
+# Why: this is a coupling that fails SILENTLY. headlong-init brings a mind up
+# with `thinkers start monolith responder`; `<name> start` in tools/persona
+# brings it back after `<name> stop`. When the two lists drift, nothing errors
+# — the dispatcher starts, status.json says "ok", the dash says "ok", and the
+# agent visibly thinks. Only the dropped thinker's job stops happening. With
+# the responder missing, messages from the human are no longer answered by the
+# thinker built to answer them promptly; they wait on the monolith's reasoning
+# loop instead, turning a few seconds of latency into minutes, and the pause/
+# resume that caused it is long forgotten by then.
+#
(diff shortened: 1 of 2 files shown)


nickjalbert:
LGTM!  Thanks!

PR #76, opened 2026-08-27
llm: pass OpenRouter's provider routing through to the request

Abeansits:
## What

`build_payload_openrouter` was a bare alias for the OpenAI builder, so the payload never carried OpenRouter's `provider` object. OpenRouter serves one model id from several hosts — a single `openai/*` id can come from OpenAI, Azure or Amazon Bedrock — and picks one per request, so without that object there is no way from here to say where a prompt may go. Anyone with a data-residency constraint can't use the OpenRouter provider at all.

Three routing knobs, each with a matching `LLM_OR_*` default so a thinker or a bridge can set one without threading a flag through every call site:

| Flag | Effect |
|------|--------|
| `--or-only SLUGS` | Pins routing to those provider slugs |
| `--or-data-collection allow\|deny` | Whether hosts that may store the prompt are allowed |
| `--or-zdr` | Zero-data-retention endpoints only |

All three apply when the effective provider is `openrouter`, whether named with `--provider`/`LLM_PROVIDER` or inferred from the model name. OpenRouter enforces them server-side — an unsatisfiable pin returns 404 rather than quietly rerouting — so this is a real constraint, not a hint.

Unpinned callers are byte-identical to before and pay no extra `jq` pass. On a non-OpenRouter provider the knobs warn on stderr and clear: the other providers are each a single host already, so a stray `LLM_OR_ZDR` in a shell profile mustn't kill every Anthropic call — but it also mustn't read as a residency guarantee, so it isn't silent.

## Changes from review

- **Rebased onto main; the line-count changes are gone.** They were written against `LOC_LIMIT: 10000` and would have lowered today's 11000.
- **`only` no longer touches `allow_fallbacks`.** You were right and my reasoning was wrong: I had them coupled on the theory that a pin which can be rerouted isn't a pin. OpenRouter's docs say `only` is the allowlist and `allow_fallbacks` is a separate retry setting, and a host outside `only` is never used either way — so the coupling bought no residency safety and cost failover *between* two hosts the caller had explicitly allowed. No replacement flag; `only` alone is what residency needs. Happy to add one if you want fallback control for its own sake.
- **Every value is normalised once, where it's read.** `1/true/yes/on` and `0/false/no/empty` for `LLM_OR_ZDR`, anything else refused — a settings file saying `LLM_OR_ZDR=true` that read as false further down would drop a retention requirement silently, which is the one failure these knobs exist to prevent. `data_collection` must be `allow` or `deny`. A pin that normalises away to nothing is now an error rather than being treated as "no pin", which would have routed anywhere.
- **An `LLM_OR_*` from the environment or a `.env` prints the effective routing once on stderr.** A constraint that decides where the prompt goes without appearing on the command line shouldn't apply invisibly. A knob typed as a flag doesn't print — you already know.
- **Help text says the settings apply to the effective provider**, named or inferred.
- **The test runs in a scratch cwd with an empty state home**, so the checkout's `.env` and `~/.headlong/.env` can't reach the results. One case writes its own `.env` to prove the layer still works and that the test owns that file.

## The jq 1.7 problem the test caught

`tests/test_llm_openrouter_routing.sh` stubs `curl` and asserts on the JSON that would have been sent. It's the first test to exercise a payload builder — every other test stubs `llm` itself, so the `jq` filters in `bin/llm` are compiled nowhere in CI.

The obvious spelling of this filter parses under jq 1.8 and is a **syntax error under jq 1.7** (`unexpected '+', expecting '}'`) — 1.7's grammar won't take a `+` chain of parenthesised `if`s as an object value. `ubuntu-latest` ships jq 1.7, so a filter only ever tried on a 1.8 dev box would be broken for every CI run and every container user. The committed form binds the sum with `as $p`, which both accept.

## Testing

- `tests/run-all.sh` — 40 passed, 0 failed
- `tests/test_llm_openrouter_routing.sh` — 20/20, and 20/20 again under jq 1.7 in a `ubuntu:24.04` container
- `shellcheck -S warning` over CI's file set — clean
- bash 3.2 parse check (macOS gate) — clean

Found running the harness against OpenRouter under a data-residency requirement.

### DIFF
--- a/bin/llm
+++ b/bin/llm
@@ -50,6 +50,33 @@ LLM_THINKING_LEVEL=""
 LLM_EFFORT=""
 LLM_RAW=0
 LLM_PROMPT=""
+# OpenRouter provider routing. OpenRouter fans a single model id out over
+# several hosts (one openai/* id can be served by OpenAI, Azure and Amazon
+# Bedrock) and picks one per request, so where a prompt is sent is a routing
+# decision rather than a model choice, and without these it is not expressible
+# from here:
+#   LLM_OR_ONLY             comma-separated provider slugs. Emits
+#                           provider.only, which OpenRouter treats as a hard
+#                           allowlist: a host outside it is never used, and an
+#                           unsatisfiable pin returns 404 rather than quietly
+#                           rerouting. Availability *within* the list is a
+#                           separate question — allow_fallbacks is OpenRouter's
+#                           retry setting, not part of the allowlist, so it is
+#                           left at its default here.
+#   LLM_OR_DATA_COLLECTION  "allow" or "deny" hosts that may store the prompt.
+#   LLM_OR_ZDR              zero-data-retention endpoints only. Intersected
+#                           with the pin rather than layered on it: a model
+#                           whose only ZDR endpoint is a host the pin excludes
+#                           has no endpoints left and 404s.
+# All three apply when the effective provider is openrouter, whether named
+# with --provider/LLM_PROVIDER or inferred from the model name.
+LLM_OR_ONLY="${LLM_OR_ONLY:-}"
+LLM_OR_DATA_COLLECTION="${LLM_OR_DATA_COLLECTION:-}"
+LLM_OR_ZDR="${LLM_OR_ZDR:-}"
+# Whether any of the three arrived from the environment or a .env layer. A
+# routing constraint nobody typed on this command line still changes where the
+# prompt may go, so it is announced once rather than applied invisibly.
+_llm_or_env="$LLM_OR_ONLY$LLM_OR_DATA_COLLECTION$LLM_OR_ZDR"
 # Token usage is captured on every call. A caller that wants this call's
 # numbers sets LLM_USAGE_FILE and gets a one-line JSON record
 # ({in_tok, out_tok, think_tok}) written there after the call (bin/shellm
@@ -156,6 +183,11 @@ Options:
   --provider NAME          Provider: anthropic, openai, gemini, openrouter,
                            opencode, opencode-go, openai-compatible, or
                            adapter (needs LLM_ADAPTER)
+  --or-only SLUGS          Pin routing to these provider slugs
+  --or-data-collection X   "allow" or "deny" hosts that may store the prompt
+  --or-zdr                 Zero-data-retention endpoints only
+                           (the three apply when the effective provider is
+                           openrouter, named or inferred from the model)
   --thinking [LEVEL]       Enable thinking (Anthropic: adaptive; Gemini: minimal/low/medium/high)
   --effort LEVEL           Thinking effort/provider output effort
   --raw                    Print raw non-streaming API response JSON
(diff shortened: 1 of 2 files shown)


nickjalbert (changes requested):
Thanks!!  The `provider` object belongs in the existing OpenRouter provider and fits [design/providers.md](https://github.com/laude-institute/headlong/blob/main/design/providers.md). The test also caught a useful jq 1.7 compatibility problem.

Requested changes from review:

> * Rebase onto main and remove the line count changes. Main already uses `LOC_LIMIT: 11000`, so `10100` would lower the limit and fail CI.
> * Normalize `LLM_OR_ZDR` once. Accept `1`, `true`, `yes`, and `on` as true. Accept `0`, `false`, `no`, and `empty` as false. Reject anything else.
> * When an `LLM_OR_*` setting comes from the environment or `.env`, print one stderr line with the effective routing constraint.
> * Do not set `allow_fallbacks: false` automatically with `only`. OpenRouter documents only as the provider allowlist and `allow_fallbacks` as a separate retry setting. Add a separate option if both behaviors are needed. Also reject a nonempty pin that becomes empty after normalization, and accept only `allow` or `deny` for `data_collection`.
> * Update the code and test for current main. Build the OpenAI payload first, then add the OpenRouter object. The help text should say the settings apply when the effective provider is `openrouter`, whether selected explicitly or inferred from the model name. In the test, assert `== "false"` and isolate the test from local `.env` files.

PR #77 contains the same stale line count changes, so please remove them there too.

Abeansits:
All five applied. Two places where I did something other than exactly what you asked:

**allow_fallbacks** — you were right and my reasoning was wrong. I had coupled it to `only` on the theory that a pin which can be rerouted isn't a pin. The docs say `only` is the allowlist and `allow_fallbacks` is a retry setting, and a host outside `only` is never used either way — so the coupling refused failover between two hosts the caller had explicitly allowed and bought no residency safety for it. I didn't add a replacement flag: `only` alone is what residency needs, and I'd rather not add surface for a behaviour nobody has asked for. Easy to add if you want fallback control on its own.

**The stderr line** only fires when the setting came from the environment or a `.env`. A knob typed as a flag stays quiet, on the grounds that you already know what you just typed.

The rest as specified. Rebased, and the line-count hunks are gone — they were written against `LOC_LIMIT: 10000` and would have lowered today's 11000, which I'd missed. `LLM_OR_ZDR` takes `1/true/yes/on` and `0/false/no/empty` and refuses anything else, `data_collection` is `allow` or `deny`, and a pin that normalises away to nothing is now an error rather than being read as "no pin" — that one would have routed anywhere, which is the opposite of the request. Help text says the settings apply to the effective provider, named or inferred.

The test runs each case in a scratch cwd with an empty state home, so the checkout's `.env` can't reach the results; one case writes its own `.env` to prove that layer still works and that the test owns the file. 20 assertions, green under jq 1.7 in a `ubuntu:24.04` container as well as 1.8 locally.

nickjalbert:
Great thanks!  LGTM!

PR #79, opened 2026-08-27
llm: add Amazon Bedrock and Mantle inference

5herlocked:
## Summary

Adds Amazon Bedrock as an inference provider for Headlong.

- Claude models use Bedrock Runtime's `InvokeModel` API.
- OpenAI `openai.gpt-*` models use the Bedrock Mantle Responses API, including streaming.
- Authentication supports Bedrock API keys, standard AWS credentials, and refreshable `AWS_PROFILE` / `credential_process` credentials.
- The installer detects Bedrock credentials, persists the selected Region and authentication configuration, and propagates credentials safely through shellm and its Docker sandbox.
- Documentation and provider-specific regression tests are included.

## Verification

- Live Claude and Mantle calls passed with both bearer-token and SigV4 authentication.
- Live profile-only calls passed using refreshed credentials from AWS CLI.
- All 32 shell test suites pass.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -32,12 +32,18 @@ trajectory.
 ## Install without a tty
 
 With no tty the installer asks nothing; every answer comes from an
-environment variable or a default. A key must be in the environment:
+environment variable or a default. A provider credential must be in the environment:
 
 ```bash
-export OPENROUTER_API_KEY=sk-or-...   # or ANTHROPIC_/OPENAI_/GEMINI_API_KEY
+export OPENROUTER_API_KEY=sk-or-...   # or another provider's *_API_KEY
 curl -fsSL https://headlong.ai/install.sh | bash
 ```
+
+Amazon Bedrock uses `AWS_BEARER_TOKEN_BEDROCK` plus `AWS_REGION` (or standard
+AWS environment credentials). `AWS_PROFILE` is also accepted and refreshes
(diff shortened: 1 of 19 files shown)


nickjalbert (changes requested):
Thanks for your work here! The live verification, provider tests, installer changes, and credential handling show that the integration was tested carefully.

Headlong now has a written [provider policy](https://github.com/laude-institute/headlong/blob/main/design/providers.md), merged in PR #78. The implementation in this PR falls outside the core provider boundary because it adds SigV4 signing, AWS profile refresh, `credential_process`, installer behavior, and sandbox credential handling to the main completion path.

Headlong now supports integrations like Bedrock through an external adapter. Operators set `LLM_PROVIDER=adapter` and `LLM_ADAPTER=/path/to/executable`. The adapter receives the request through stdin and documented flags, then writes the response to stdout. `bin/llm` continues to manage the deadline, child process cleanup, health marker, and usage ledger.

Bedrock adapter can use the AWS CLI or an SDK and can manage AWS credentials without adding those dependencies to core. The adapter can support both Claude through `InvokeModel` and OpenAI models through Mantle. It can also keep the credential refresh behavior you implemented without reimplementing SigV4 in Bash.

I would welcome a reworked PR that moves the Bedrock implementation behind the adapter interface.

5herlocked:
Will do - makes sense to move providers into an adapter pattern.

PR #82, opened 2026-08-28
shellm: a sibling directory is not a mounted one

### LINKED ISSUE #81: shellm: a directory sharing a mounted path's prefix is silently never mounted
## The bug

`docker_setup` decides whether a directory is already covered by an existing bind mount with a bare prefix test:

https://github.com/laude-institute/headlong/blob/f8e1f228ef84/bin/shellm#L852-L856

A path is a textual prefix of its siblings, so `/x/workdirs` reads as covering `/x/workdirs-scratch`. The sibling is judged already mounted and gets no `-v`. Docker still receives `-w` on that path, so it creates an empty root-owned directory there and the run continues: the operator's files are invisible to the agent, anything the agent writes is discarded when the container goes away, and nothing reports an error.

Four comparisons have this shape (lines 827, 852, 855, 870 at f8e1f22). Two others in the same function, at 838 and 900, already use `== "$d" || == "$d"/*`, so this looks like drift rather than a decision.

## Reproducing it

With a workdir whose path starts with the workdirs dir but is not under it:

```bash
base=$(mktemp -d)
mkdir -p "$base/workdirs" "$base/workdirs-scratch"
echo hello > "$base/workdirs-scratch/note.txt"

SHELLM_WORKDIRS_DIR="$base/workdirs" shellm --workdir "$base/workdirs-scratch" 'ls -la; cat note.txt'
```

`docker inspect` on the resulting container lists a mount for `$base/workdirs` and none for the workdir, and the agent sees an empty directory rather than `note.txt`. The same thing happens to a `-f` file's directory and a `--var` directory when either is a prefix-sharing sibling of an already mounted path.

## It also stops envs from being reused

`docker_container_can_reach_path` compares against real mount destinations, so it correctly concludes the container cannot reach the workdir and `env_resolve` skips the env:

```
Skipping env myenv because it cannot reach workdir /tmp/.../workdirs-scratch
```

`docker_setup` then builds a new container with the same missing mount, and the next run repeats it. Three runs with the same `--workdir` produced three containers, where an ordinary workdir produced one and reused it twice. Containers accumulate for as long as that workdir is used.

## A path boundary alone does not fix it

Worth flagging because it is the obvious patch. The predicate also never normalizes, and `--workdir "$SHELLM_WORKDIRS_DIR/../scratch"` genuinely is `"$runs_mount"/*` as a string, so it stays unmounted after the comparison is given a boundary. I checked this against a build with the boundary applied.

These are container paths rather than host paths: inside the container a path is reachable only if it sits under a mount destination, and docker normalizes `..` in both `-w` and destinations (`-w /a/../b` lands on `/a/b`, and a destination containing `..` is stored normalized). So the comparison wants lexical normalization, and specifically not symlink resolution: resolving would judge the host source, and a workdir symlinked into the workdirs dir but mounted at its own link path would start reading as covered by a mount that does not contain it.

## Environment

docker 29.1.3, Ubuntu, bin/shellm at f8e1f22.

I have a fix with tests and will open a PR against this.

beardthelion:
Fixes #81.

`docker_setup` decided whether a directory was already covered by an existing bind mount with a bare prefix test, `[[ "$abs_dir" == "$d"* ]]`. A path is a textual prefix of its siblings, so `/x/workdirs` read as covering `/x/workdirs-scratch` and the sibling got no `-v`. Docker still received `-w` on that path and created an empty directory there, so the run continued with the operator's files invisible and the agent's writes discarded at teardown, with no error on any surface.

It also decided container reuse. `docker_container_can_reach_path` compares against real mount destinations, so it correctly reported that the container could not reach the workdir and forced a new one, which `docker_setup` then built with the same missing mount. Three runs with the same `--workdir` produced three containers where an ordinary workdir produced one and reused it twice.

## Why a path boundary is not the whole fix

The obvious patch is `== "$d" || == "$d"/*`, which is what lines 838 and 900 already did. It closes the sibling case and leaves the class open: the predicate also never normalized, and `--workdir "$SHELLM_WORKDIRS_DIR/../scratch"` genuinely is `"$runs_mount"/*` as a string, so it stays unmounted. I ran that against a boundary-only build to be sure rather than assuming it.

The comparison also wants normalization rather than resolution, which is the opposite of the call made for the compose gate in #47. These are container paths, not host paths: inside the container a path is reachable only if it sits under a mount destination, and docker normalizes `..` in both `-w` and destinations (`-w /a/../b` lands on `/a/b`, and a destination containing `..` is stored normalized). Resolving symlinks would judge the host source instead, so a workdir symlinked into the workdirs dir but mounted at its own link path would begin reading as covered by a mount that does not contain it.

So both halves now go through one `_path_covered` that normalizes lexically and requires a boundary, used at all seven sites: the four that were wrong, the two that already had the boundary right, and the reuse check that had drifted from them.

## Verification

Test-first, with the seam extracted carrying the old behavior first so the red run was a real failure rather than `command not found`: 10 passed and 6 failed with the wrong values printed, then 27 passed after the change.

Against a real daemon (docker 29.1.3), asserting on the mount list `docker_setup` produced:

| case | expected | before | after |
| --- | --- | --- | --- |
| workdir is a prefix-sharing sibling | mounted | missing | mounted |
| workdir reached through `..` | mounted | missing | mounted |
| `-f` file's dir is a prefix-sharing sibling | mounted | missing | mounted |
| `--var` dir is a prefix-sharing sibling | mounted | missing | mounted |
| workdir inside the workdirs dir | reachable | ok | ok |
| workdir plainly outside it | mounted | ok | ok |
| identity dir is a prefix-sharing sibling | mounted | ok | ok |
| symlink target dir is a prefix-sharing sibling | mounted | ok | ok |
| identity dir, `-f` dir, `--var` dir, symlink target already inside the workdir | no second mount | ok | ok |

Three runs with the same `--workdir` now create one container and reuse it, for both the sibling and the `..` spelling.

`tests/run-all.sh` passes, the shellcheck gate is clean, and the new test passes under bash 3.2.57 as well as bash 5. I have not run the macOS job itself; the change shells out to nothing, so it is pure bash 3.2 syntax, but that job is the real gate.

The unit test covers the predicate directly, including the degenerate inputs (empty, `/`, repeated slashes, `..` climbing past the root) and segments containing glob and bracket characters, since the helper avoids word splitting on purpose.

### DIFF
--- a/bin/shellm
+++ b/bin/shellm
@@ -758,6 +758,48 @@ env_metadata_mismatch_message() {
         "$env_name" "$stored_access" "$SHELLM_DOCKER_ACCESS"
 }
 
+# Collapse an absolute path lexically: drop empty and `.` segments, pop one
+# level per `..`, and emit no trailing slash. Purely textual, and deliberately
+# not realpath: these paths are compared as CONTAINER paths, and docker
+# normalizes `..` the same way in both -w and a mount destination (`-w /a/../b`
+# lands on /a/b). Resolving symlinks here would judge the host source instead,
+# so a workdir symlinked into the runs dir but mounted at its own link path
+# would read as covered by a mount that does not contain it.
+#
+# No forks and no word splitting: bin/shellm also runs on macOS bash 3.2, and
+# splitting on IFS=/ would glob a segment containing `*`.
+_norm_path() {
+    local rest="$1" out="" seg
+    while [[ -n "$rest" ]]; do
+        seg="${rest%%/*}"
+        if [[ "$seg" == "$rest" ]]; then rest=""; else rest="${rest#*/}"; fi
+        case "$seg" in
+            ""|.) ;;
+            ..)   out="${out%/*}" ;;
+            *)    out="$out/$seg" ;;
+        esac
+    done
+    printf '%s\n' "${out:-/}"
+}
+
+# True when `path` is already covered by an existing bind mount rooted at
+# `root`. Both are container-side paths: these decide whether docker_setup
+# still needs a -v for a directory, and inside the container a path is
+# reachable only if it sits under a mount destination.
+#
+# Normalize before comparing, and require a path boundary. A bare prefix test
+# reads a sibling as a child (`/x/workdirs` "covering" `/x/workdirs-scratch`),
+# which cost the sibling its -v while docker still made the -w directory, so
+# the run continued against an empty directory with no error anywhere.
+_path_covered() {
+    local path root
+    path=$(_norm_path "$1")
+    root=$(_norm_path "$2")
+    # ${root%/} so a root of "/" yields the pattern /* rather than //*, which
+    # matches nothing and would leave every path uncovered by the filesystem root.
+    [[ "$path" == "$root" || "$path" == "${root%/}"/* ]]
+}
+
 docker_setup_die() {
     local message="$1"
     if [[ -n "${_SHELLM_CONTAINER:-}" ]]; then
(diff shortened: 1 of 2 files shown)


nickjalbert:
🚢

PR #84, opened 2026-08-28
telegram: send chat files as documents/photos

headlong42:
This slice:
- detects file-bearing mind-log steps (`filename` / PNG / SVG sniff)
- adds `Bot.send_document` and `Bot.send_photo` (multipart)
- routes images to `sendPhoto` and everything else to `sendDocument`
- keeps the existing source=chat delivery guard

Intentionally not in this PR: identity-local responder dest-history stub, notes, secrets, web build artifacts.

Pushed from `headlong42/headlong` because `headlong42` has pull-only on `laude-institute/headlong` (direct origin push 403).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -274,7 +274,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 5
     env:
-      LOC_LIMIT: 11000
+      LOC_LIMIT: 11500
     steps:
       - uses: actions/checkout@v5
       - name: Install cloc
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 [![CI](https://github.com/laude-institute/headlong/actions/workflows/ci.yml/badge.svg)](https://github.com/laude-institute/headlong/actions/workflows/ci.yml)
 
 **Headlong** is an open source agent microharness, a complete agent harness
-with a core of about 10K lines of Bash.
+with a core of about 11K lines of Bash.
 
 [Launch post](https://www.laude.org/updates/headlong-a-microharness-for-persistent-agents) |
 [Announcement](https://x.com/andykonwinski/status/2091990178638496195)
@@ -158,7 +158,7 @@ Headlong also gives an agent a few convenience tools, such as a way to
 distill and codify its experience (`mem`) and a way to save and reuse
 procedures for specialized tasks (`skills`). The core is the tools the
 running mind executes, the executables in `bin/` plus the thought
-processes in `thinkers/`, and it comes to about 10K lines by cloc's count. A
+processes in `thinkers/`, and it comes to about 11K lines by cloc's count (capped at 11.5K). A
 harness this small can be read end to end, and it is easy to modify and
 experiment with.
 
--- a/bin/chat
+++ b/bin/chat
@@ -214,15 +214,22 @@ cmd_send_file() {
 
     local filename
     filename=$(basename "$filepath")
-    local content
-    content=$(cat "$filepath")
+    # Binary-safe: JSON cannot hold raw bytes, so stamp standard base64.
+    local content_b64
+    content_b64=$(python3 -c 'import base64,sys; print(base64.b64encode(open(sys.argv[1],"rb").read()).decode("ascii"))' "$filepath") \
+        || die "failed to base64-encode $filepath"
 
-    printf '{"type":"message","filename":%s,"content":%s,"from":%s,"to":%s,"source":"chat"}' \
-        "$(printf '%s' "$filename" | jq -Rsa .)" \
-        "$(printf '%s' "$content" | jq -Rsa .)" \
-        "$(printf '%s' "$from" | jq -Rsa .)" \
-        "$(printf '%s' "$to" | jq -Rsa .)" \
-        | _chat_append
+    python3 -c 'import json,sys
+filename, content_b64, frm, to = sys.argv[1:5]
+print(json.dumps({
+    "type": "message",
+    "filename": filename,
+    "content_b64": content_b64,
+    "from": frm,
+    "to": to,
+    "source": "chat",
+}, separators=(",", ":")))
+' "$filename" "$content_b64" "$from" "$to" | _chat_append
 
     printf 'File sent: %s\n' "$filename" >&2
 }
--- a/bin/chat
+++ b/bin/chat
@@ -219,10 +219,11 @@ cmd_send_file() {
     # (same class of leak as API keys on curl command lines).
     base64 -w 0 < "$filepath" | tr -d '\n' | jq -nc \
         --arg filename "$filename" \
+        --arg content "[file: $filename]" \
         --arg from "$from" \
         --arg to "$to" \
         --rawfile content_b64 /dev/stdin \
-        '{type:"message", filename:$filename, content_b64:$content_b64, from:$from, to:$to, source:"chat"}' \
+        '{type:"message", content:$content, filename:$filename, content_b64:$content_b64, from:$from, to:$to, source:"chat"}' \
         | _chat_append \
         || die "failed to encode/append $filepath"
 
--- a/bin/chat
+++ b/bin/chat
@@ -226,20 +226,56 @@ cmd_send_file() {
 
     local filename
     filename=$(basename "$filepath")
-    # Binary-safe: JSON cannot hold raw bytes, so stamp standard base64.
-    # Read the encoding via jq --rawfile so the bytes never sit on argv
-    # (same class of leak as API keys on curl command lines).
-    base64 -w 0 < "$filepath" | tr -d '\n' | jq -nc \
-        --arg filename "$filename" \
-        --arg content "[file: $filename]" \
-        --arg from "$from" \
-        --arg to "$to" \
-        --rawfile content_b64 /dev/stdin \
-        '{type:"message", content:$content, filename:$filename, content_b64:$content_b64, from:$from, to:$to, source:"chat"}' \
-        | _chat_append \
-        || die "failed to encode/append $filepath"
+    # Binary-safe: JSON cannot hold raw bytes, so stamp standard base64
+    # via a temp file (never argv). Text files also keep their body in
+    # `content` so Slack and web chat still deliver the file text; binary
+    # files use a short marker because `content` cannot hold raw bytes.
+    local b64tmp
+    b64tmp=$(mktemp) || die "mktemp failed"
+    if ! base64 -w 0 < "$filepath" | tr -d '\n' > "$b64tmp"; then
+        rm -f "$b64tmp"
+        die "failed to encode $filepath"
+    fi
+
+    local jq_rc=0
+    if _chat_is_text_file "$filepath"; then
+        jq -nc \
+            --arg filename "$filename" \
+            --arg from "$from" \
+            --arg to "$to" \
+            --rawfile content "$filepath" \
+            --rawfile content_b64 "$b64tmp" \
+            '{type:"message", content:($content | sub("\n+$";"")), filename:$filename, content_b64:$content_b64, from:$from, to:$to, source:"chat"}' \
+            | _chat_append || jq_rc=$?
+    else
+        jq -nc \
+            --arg filename "$filename" \
+            --arg content "[file: $filename]" \
+            --arg from "$from" \
+            --arg to "$to" \
+            --rawfile content_b64 "$b64tmp" \
+            '{type:"message", content:$content, filename:$filename, content_b64:$content_b64, from:$from, to:$to, source:"chat"}' \
+            | _chat_append || jq_rc=$?
+    fi
+    rm -f "$b64tmp"
+    [[ "$jq_rc" -eq 0 ]] || die "failed to encode/append $filepath"
 
     printf 'File sent: %s\n' "$filename" >&2
+}
+
+# UTF-8, no NUL: the historical `content=$(cat ...)` shape Slack/web still read.
+# JSON strings cannot hold NUL; jq --rawfile requires UTF-8.
+_chat_is_text_file() {
+    local f="$1" n stripped
+    n=$(wc -c < "$f")
+    n=${n//[[:space:]]/}
+    stripped=$(tr -d '\0' < "$f" | wc -c)
+    stripped=${stripped//[[:space:]]/}
+    [[ "$n" -eq "$stripped" ]] || return 1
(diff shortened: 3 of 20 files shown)


nickjalbert (changes requested):
Hi @headlong42!  From the review:

> 
> The file branch is placed correctly: it runs after the `source == "chat"` guard, the telegram-name/identity checks, the allowlist check, and dedupe, and `file_payload` reads only fields off the step and reduces filenames with `Path(...).name`, so it cannot be steered into reading files off the box. Reading the checkout, five reviewers still found the slice not ready:
> 
> 1. `bin/chat` (`cmd_send_file`, ~205-209) + `telegram/src/headlong_telegram/filepayload.py` — decide and implement how binary file content reaches the trajectory. Today `content=$(cat ...)` plus `jq -Rsa` drops NULs and replaces invalid UTF-8, and `mindlog` (`json.loads`) always yields `str`, so the `isinstance(content, (bytes, bytearray))` PNG-magic path is unreachable in production and a real PNG arrives corrupted. Either add an explicit base64-encoded field that the bridge decodes, or restrict this slice to text files and delete the bytes branches plus the tests that inject `bytes` steps.
> 2. `telegram/src/headlong_telegram/outbound.py` (file branch) — on `ApiError` from `send_photo`, retry the same payload with `send_document` before consuming the step, and widen the `except` to `(ApiError, httpx.HTTPError)`. `mindlog.follow` persists the cursor before yielding, so today a rejected or timed-out upload is lost with no chat-visible trace, and a non-`ApiError` transport failure ends the outbound daemon thread.
> 3. `telegram/src/headlong_telegram/outbound.py` — gate the file branch on the delivery target actually being authorized (e.g. require `conv.user == conv.chat` for private routes, or authorize `conv.chat`). The allowlist checks `conv.user` while `send_photo`/`send_document` target `conv.chat`, and `naming.decode` accepts `telegram-<approved-user>-<other-chat>`. The mismatch is pre-existing on the text path, so the maintainer may prefer to fix it repo-wide; either way, uploads should not ship on top of it unguarded.
> 4. `telegram/src/headlong_telegram/filepayload.py` — key file detection on the explicit `filename` stamp that `chat send-file` writes; drop the `file` alias, the `<svg`-prefix sniff, and the `FILE_TYPES`/`typed` fallback (unreachable, since the caller already requires `type == "message"`). Also truncate captions to Telegram's 1024-character media-caption limit rather than 2000, since an over-long caption currently drops the whole file.
> 5. `telegram/tests/test_outbound.py` — restore `test_recent_posts_dedupe_window`. It was the only coverage of `RecentPosts`, which is unchanged by this PR and is one of the guards the new branch depends on.
> 

Definitely interested in this feature but please fix the above before merge.

nickjalbert:
@headlong42 Thanks for the iteration. The route check against the allowlist, the restored dedupe test, the httpx error handling, and the 1024 caption limit all look right.

The fix commit introduced problems of its own:

- bin/chat now passes the whole base64 file as a single command argument. File contents become visible in the process list, and a file near the OS argument size limit fails outright. We just removed this exact pattern from the repo in #83. Stream the base64 over stdin or through a temp file instead.
- bin/chat now runs inline python3. The core bin/ path is bash plus standard tools by design. Build the JSON with jq and the system base64 command instead.
- outbound.py contains the same file dispatch block three times (around lines 81, 103, and 125). Only the first copy can ever run. Keep one and delete the other two.
- Photo detection only recognizes PNG magic, so a JPEG always arrives as a document. Either recognize JPEG too or tell me that sending everything as a document is intended.

Still open from the first review:

- File content and captions bypass strip_leaked_command. Run them through the same leak filter as text messages.
- deploy/SECURITY.md still says the bridge is text only and drops media. Update it to describe file delivery.
- When the final send fails, the failure is only logged and the cursor has already advanced, so the file is silently lost. At minimum, fall back to a text message saying the delivery failed.
- The "file" alias for "filename" can still reclassify an ordinary message as an upload. Accept only the canonical key.

let me know when you make another pass.

headlong42:
Made another pass. Remaining items from the last review:

- `bin/chat send-file` no longer puts file bytes on argv. It pipes `base64` into `jq --rawfile` (no inline python).
- JPEG magic is treated like PNG (`sendPhoto`); other binaries still go as documents.
- Duplicated outbound file-send block is gone.
- Canonical `filename` only — a `file` field no longer reclassifies an ordinary message.
- Text file bodies and captions run through `strip_leaked_command`.
- If `sendPhoto`/`sendDocument` fails after the cursor has already advanced, we fall back to a short text notice (`(failed to deliver file …)`) so the file isn't silently lost.
- `deploy/SECURITY.md` now distinguishes inbound media (still dropped) from outbound `chat send-file` uploads.

Latest commit: d423537. telegram filepayload/outbound/tgfmt tests: 28 passed.

nickjalbert:
From review:

>
> 1. `bin/chat` (`cmd_send_file`) + readers: file steps now carry only `content_b64`, and `slack/src/headlong_slack/outbound.py:74`, `web/src/headlong_web/trajectory.py:318`, `web/src/headlong_web/push.py:117`, `tui/headlong/src/main.rs:473`, and `bin/chat:231` (`cmd_history`) all skip steps with empty `content` — `chat send-file --to <slack-name>` is a silent no-op and Telegram sends vanish from history/dashboards. Please keep a bounded human-readable `content` (e.g. `[file: fig.png]`) alongside `content_b64`.
> 2. `thinkers/_lib/common.sh:~253` (and check `bin/context`): the recent-stream filter truncates only `.content`, so the full base64 payload is interpolated into `thinkers/monolith/step` route prompts — unbounded token cost and possible context overflow. Elide `content_b64` from model-facing renderers and render filename metadata instead.
> 3. `tests/`: add a bash test (`SHELLM_ENV=local`) that runs `chat send-file` on a fixture containing NUL bytes and newlines and asserts `content_b64` decodes back to the original bytes and `filename` is the basename. The producer pipeline is currently executed by no test on any platform.
> 4. `bin/chat` (`cmd_send_file`): reject files above a small cap (well under Telegram's 50MB bot limit). There is no size guard, the log inflates 4/3, and the repo has a documented 590MB-log OOM.
> 5. `bin/chat` / `telegram/src/headlong_telegram/filepayload.py`: an empty file, or a `content_b64` that fails strict decode, yields "File sent" to the agent and nothing delivered or logged. Fail loudly in one of the two places.
> 6. `design/trajectory_spec.md:391-394`: the documented file-step schema still requires `content` and does not mention `content_b64`. Update it to match whatever shape (1) settles on.
> 

Thanks for the continued work on this @headlong42 — the Telegram side is in good shape now, and the review confirmed the base64 round-trip and the group-chat guard both hold up.

On (1): please keep a short human-readable `content` alongside `content_b64`, e.g. `[file: fig.png]`. That's a one-line change in `bin/chat` and every existing reader keeps working. Moving the readers onto `content_b64` properly is a bigger change we may do later, but it's out of scope for this PR. Note the Telegram side already prefers `content_b64`, so no Python change is needed.

For (6), document the `content` + `content_b64` pair as the file-step shape.

headlong42:
Pushed 34f8286 for the remaining notes:

- `bin/chat send-file` now stamps a short human-readable `content` (e.g. `[file: fig.png]`) alongside `content_b64`.
- `design/trajectory_spec.md` documents that pair as the file-step shape.

Telegram side unchanged (already prefers `content_b64`).

nickjalbert:
Thanks @headlong42, the two fixes you pushed look good but you missed several notes though.   

>
> Confirmed the content marker and spec doc — the reader-visibility blocker is resolved. 
>
> Still open from the review: items 2, 3, 4, and 5. In detail:
>
> 2. `thinkers/_lib/common.sh:~253` (and check `bin/context`): the recent-stream filter truncates only `.content`, so the full base64 payload is interpolated into `thinkers/monolith/step` route prompts — unbounded token cost and possible context overflow. Elide `content_b64` from model-facing renderers and render filename metadata instead.
> 3. `tests/`: add a bash test (`SHELLM_ENV=local`) that runs `chat send-file` on a fixture containing NUL bytes and newlines and asserts `content_b64` decodes back to the original bytes and `filename` is the basename. The producer pipeline is currently executed by no test on any platform.
> 4. `bin/chat` (`cmd_send_file`): reject files above a small cap (well under Telegram's 50MB bot limit). There is no size guard, the log inflates 4/3, and the repo has a documented 590MB-log OOM.
> 5. `bin/chat` / `telegram/src/headlong_telegram/filepayload.py`: an empty file, or a `content_b64` that fails strict decode, yields "File sent" to the agent and nothing delivered or logged. Fail loudly in one of the two places.
> 


This will be mergeable after those four fixes.

nickjalbert (changes requested):
Thanks!  From review:

 > 1. **Preserve `chat send-file` behavior for other transports** ([bin/chat:238](https://github.com/laude-institute/headlong/blob/2c89d2e8f7dbc0c9eebfc31542a07516ac3b1fb6/bin/chat#L238))
  >
  > `chat send-file` is a shared Headlong command. Before this change, a text file's contents were stored in `content`, so Slack and web chat delivered those contents. The new step stores only `[file: name]` in `content`, while only Telegram decodes `content_b64`. Slack and web chat now deliver the marker instead of the file contents.
  >
  > Preserve the existing text file behavior for other readers, or update every supported transport to consume the new file payload. Add a regression test showing that a text file sent through a non Telegram route still delivers its contents.
  > 
  > 2. **Remove file bytes from the dashboard API response** ([web/src/headlong_web/trajectory.py:239](https://github.com/laude-institute/headlong/blob/2c89d2e8f7dbc0c9eebfc31542a07516ac3b1fb6/web/src/headlong_web/trajectory.py#L239))
  >
  > The dashboard normalizer copies the complete step into `raw`, including `content_b64`, and the mind log endpoint sends that object to the browser. A 1 MiB test file produced a 1,398,488 byte response. A file at the new 10 MiB limit adds about 13.3 MiB to each response and remains in browser memory even though the dashboard displays only the short marker.
  >
  > Strip `content_b64` from the dashboard wire representation during both normal ingestion and raw rehydration. Add a web API test showing that `filename` and the marker remain visible while `content_b64` is absent.
  > 
  > 3. **Apply duplicate suppression to file uploads** ([telegram/src/headlong_telegram/outbound.py:81](https://github.com/laude-institute/headlong/blob/2c89d2e8f7dbc0c9eebfc31542a07516ac3b1fb6/telegram/src/headlong_telegram/outbound.py#L81))
  >
  > The file branch returns before the existing `RecentPosts` check at line 121. Two identical file steps therefore produce two Telegram uploads. I reproduced this with two steps containing the same filename and bytes.
  >
  > Build a bounded file signature, such as the filename plus a hash of the decoded bytes, and pass it through duplicate suppression before uploading. Test that identical files are suppressed while different contents with the same filename are both delivered.
  > 
  > ## Other comments
  > 
  > 1. **Nit: Remove blank lines at end of the new Python tests** ([telegram/tests/test_filepayload.py:111](https://github.com/laude-institute/headlong/blob/2c89d2e8f7dbc0c9eebfc31542a07516ac3b1fb6/telegram/tests/test_filepayload.py#L111))
  >
  > `git diff --check` reports extra blank lines at the ends of `test_filepayload.py` and `test_outbound.py`.

nickjalbert (changes requested):
## Must fix before merge

> 1. **Fix line count gate** ([bin/chat:201](https://github.com/laude-institute/headlong/blob/c6476094733b9433fb81c45d73977d9b3b145708/bin/chat#L201))
>
> The merge is blocked by the failing [core line count check](https://github.com/laude-institute/headlong/actions/runs/33778979599/job/100727690839). The merge base has 10,980 code lines in `bin/` and `thinkers/`, while this head has 11,031. Current `main` has 10,996, and the automatic merge has 11,047 against the 11,000 line limit.
>
> Please raise the cap to 11.5k and update the README (can be hand-weavey). Verify the result with `cloc --quiet --json bin thinkers | jq -r '.SUM.code'` and a green `cloc bin/ thinkers/ under limit` job.
>
> 2. **Reject invalid file content without stopping Telegram replies** ([telegram/src/headlong_telegram/filepayload.py:46](https://github.com/laude-institute/headlong/blob/c6476094733b9433fb81c45d73977d9b3b145708/telegram/src/headlong_telegram/filepayload.py#L46), [telegram/src/headlong_telegram/outbound.py:91](https://github.com/laude-institute/headlong/blob/c6476094733b9433fb81c45d73977d9b3b145708/telegram/src/headlong_telegram/outbound.py#L91))
>
> `_content` returns any nonempty JSON value from a legacy file step, although the rest of the code accepts only text or bytes. A step such as `{"filename":"bad.bin","content":{"x":"y"}}` reaches `file_signature`, where `bytes(content)` raises `TypeError` before the send error handler. The outbound worker is a daemon thread, so it stops delivering later replies while the inbound bridge process stays alive. I reproduced the exception and confirmed that the next valid message was not sent.
>
> Please accept only string content for the legacy JSON form, treat every other value as a decode error, and continue to the existing failure notice. Add a regression test with an invalid file step followed by a valid message, and assert that the notice and the later message are both delivered. Verify with the full Telegram test suite.

nickjalbert (changes requested):
One more reqeust from review:

## Must fix before merge

> 1. **Remove encoded file bytes from slim exports** ([bin/chat:247](https://github.com/laude-institute/headlong/blob/21c62ba7264faf8603a927d3b1ea76f56a227d5c/bin/chat#L247))
>
> `chat send-file` now stores the complete file in `content_b64`. The slim filter in [`tools/identity`](https://github.com/laude-institute/headlong/blob/21c62ba7264faf8603a927d3b1ea76f56a227d5c/tools/identity#L493) and the separate filter in [`deploy/scripts/audel-export`](https://github.com/laude-institute/headlong/blob/21c62ba7264faf8603a927d3b1ea76f56a227d5c/deploy/scripts/audel-export#L121) preserve that field. Their later text redaction cannot find a credential inside base64. I reproduced the problem with a fake OpenRouter key inside `secret.bin`. `identity export --slim` kept `content_b64`, and decoding it returned the key unchanged. Anyone sharing that export can therefore disclose file contents that the export redactor was supposed to remove. The field also defeats the size reduction expected from a slim export.
>
> Delete `content_b64` in both slim transforms while keeping `filename` and the readable `content` body or marker. Add a case to `tests/test_identity_export_import.sh` that places a file step with a credential shaped payload in the trajectory, runs the real slim exporter, and proves that the field and decoded payload are absent. Verify the separate Audel export transform as well.

nickjalbert:
LGTM!  Thanks!

PR #87, opened 2026-08-28
shellm: thinking on stderr is not an llm failure

### LINKED ISSUE #86: shellm: a model that spends its whole budget on thinking kills the run instead of retrying
## The bug

`llm --thinking` streams the model's reasoning to stderr, and a call whose token budget runs out during that reasoning comes back as a 200 with no visible text: stderr carries the thinking plus `llm: warning: output truncated at max_tokens=...`, stdout is empty, exit status 0.

`call_llm` reads that as a failure:

https://github.com/laude-institute/headlong/blob/ed2dac6cfe9a/bin/shellm#L1401-L1407

So the run dies, with the model's reasoning printed as the error message.

That is the exact case `run_loop`'s empty-response retry was written for: feed the captured thinking back as assistant context and let the model continue from where it stopped.

https://github.com/laude-institute/headlong/blob/ed2dac6cfe9a/bin/shellm#L2294-L2297

The retry needs an empty response *and* non-empty captured thinking, and the thinking is the stderr the guard dies on, so the feed-back branch at line 2307 is unreachable. The only way past the guard is a provider that emits whitespace-only stdout: the `-s` test then passes and `$( )` strips it back to empty.

The guard is older than the exit-code check above it. It landed in c3ef334 (2026-06-01, "surface llm errors to user"), when `call_llm` captured llm's exit code but never looked at it. 58cc447 added the real check two months later, and every error path in `bin/llm` goes through its `die()` (line 83) and exits 1, so exit 0 with content on stderr is thinking, a truncation warning (line 794) or a transient-retry notice, never an error.

## Reproducing it

A stub `llm` that imitates the truncated-during-thinking response:

```bash
mkdir -p /tmp/stub/bin
cat > /tmp/stub/bin/llm <<'EOF'
#!/usr/bin/env bash
case "$*" in *--thinking*) ;; *) printf '{"summary":"stub"}\n'; exit 0 ;; esac
n=$(cat /tmp/stub/n 2>/dev/null || echo 0); echo $((n+1)) > /tmp/stub/n
if [ "$n" -eq 0 ]; then printf 'Thinking: budget spent on reasoning\n' >&2; exit 0; fi
printf 'Thinking: now I answer\n' >&2
printf 'Here we go.\n\n```bash\nFINAL="OK"\n```\n'
EOF
chmod +x /tmp/stub/bin/llm; rm -f /tmp/stub/n

PATH=/tmp/stub/bin:$PATH SHELLM_MODEL=stub SHELLM_FAST_MODEL=stub \
  shellm --max-iterations 2 "do the thing"
```

```
▶ Iteration 1/2 — calling stub...
shellm: error: llm failed: Thinking: budget spent on reasoning
```

The retry never runs; the second call never happens. Expected is the retry firing and the run continuing.

It also reproduces against a real `bin/llm`, with `SHELLM_API_URL` pointed at a local endpoint that returns a thinking-only Anthropic stream ending in `"stop_reason": "max_tokens"`. `bin/llm` exits 0 with empty stdout there, and `shellm` dies on it.

## Scope

Any provider and any model, whenever a call spends its budget on reasoning before emitting text. Reasoning-heavy models at a high `--effort` are where it shows up. A run that should have recovered by itself ends instead, and the operator gets the model's chain of thought as an error message.

Same shape without `--thinking`: a call that legitimately returns nothing but writes anything to stderr, a transient-retry notice included, dies rather than retrying.

## Environment

Ubuntu, bin/shellm at ed2dac6.

I have a fix with tests and will open a PR against this.

beardthelion:
Fixes #86.

`llm --thinking` puts the model's reasoning on stderr, and a call that spends its whole token budget there returns 200 with nothing on stdout and exits 0. `call_llm` treated "no stdout, something on stderr" as a failure and died, printing the reasoning as the error.

That is what `run_loop`'s empty-response retry exists for. The retry needs an empty response together with non-empty captured thinking, and the thinking is the stderr the guard died on, so the feed-back branch was unreachable: the only way past the guard was a provider emitting whitespace-only stdout, which `$( )` then strips back to empty.

The guard predates the exit-code check above it (c3ef334, before `call_llm` looked at llm's exit code at all; 58cc447 added the check two months later). Every error path in `bin/llm` goes through its `die()` and exits 1, so exit 0 with content on stderr is thinking, a truncation warning or a transient-retry notice, and the check above already covers the failures. Removing the guard leaves the empty-response policy where it was written, in `run_loop`.

## Verification

`tests/test_shellm_empty_response_retry.sh` extracts `call_llm` and drives it against a stub `llm`, in both quiet modes: thinking with no visible text returns empty and keeps the thinking for the retry, whitespace-only and silent responses return empty, a normal answer is unaffected, and a non-zero llm exit still fails the call naming the exit code. 14 assertions, RED before the change. Restoring the guard turns the two thinking-only assertions red; deleting the exit-code check turns the four failure assertions red.

The test runs `call_llm` inside a command substitution the way `run_loop` does. That nesting is not cosmetic: bash does not carry `errexit` into the non-quiet branch's pipeline subshell from there, which is what lets that branch record llm's exit code.

End to end through the real stack, with `SHELLM_API_URL` pointed at a local endpoint returning a thinking-only Anthropic stream that ends in `"stop_reason": "max_tokens"`:

| | exit | provider calls | result |
|---|---|---|---|
| before | 1 | 2 | `shellm: error: llm failed: <the reasoning>` |
| after | 0 | 3 | retry fires, second call answers, run completes |

Eight branches were run end to end through `bin/shellm` before and after, on bash 5.2 and on bash 3.2.57: thinking-only then answer, silent-empty then answer, always-thinking-only against a retry cap, always-silent against a cap, a non-zero llm exit, a normal answer, quiet mode, and whitespace-only stdout. Only the three cases that reach the guard change; the terminal "after N retries" deaths and the exit-code death are unchanged.

Also run locally: `tests/run-all.sh` 36/36 on bash 5.2 and on bash 3.2.57 (built from source for this), `bash -n` and the bash-4 construct lint over all 71 product scripts under 3.2, `shellcheck -S warning` over the tree, pytest for web, slack and telegram, `tests/smoke_install.sh` in a bare `ubuntu:24.04` container, and the cargo and viewer checks.

### DIFF
--- a/bin/shellm
+++ b/bin/shellm
@@ -1398,13 +1398,13 @@ call_llm() {
         die "llm failed (exit $llm_exit): $_llm_err"
     fi
 
-    # If llm produced no output, surface any error from stderr
-    if [[ ! -s "$text_file" && -s "$stderr_file" ]]; then
-        local _llm_err
-        _llm_err=$(cat "$stderr_file")
-        rm -f "$text_file" "$stderr_file"
-        die "llm failed: $_llm_err"
-    fi
+    # No visible text is not a failure. `llm --thinking` streams reasoning to
+    # stderr, and a run that spends its whole token budget there returns 200,
+    # exits 0 and emits nothing on stdout, which is exactly what run_loop's
+    # empty-response retry feeds back as assistant context. Dying on a
+    # non-empty stderr here made that retry unreachable and printed the
+    # model's reasoning as an error. Real failures exit non-zero and are
+    # caught above.
 
     cat "$text_file"
     rm -f "$text_file" "$stderr_file"
--- a/tests/test_shellm_empty_response_retry.sh
+++ b/tests/test_shellm_empty_response_retry.sh
@@ -1,0 +1,108 @@
+#!/usr/bin/env bash
+# tests/test_shellm_empty_response_retry.sh — call_llm only fails the call
+# when llm actually failed.
+#
+# Usage: tests/test_shellm_empty_response_retry.sh
+#
+# `llm --thinking` streams the model's reasoning to stderr and, when the
+# token budget runs out during that reasoning, returns 200 with no visible
+# text and exits 0. run_loop's empty-response retry exists for exactly that
+# case (feed the captured thinking back as assistant context and continue),
+# so call_llm must hand it back an empty response instead of dying on the
+# thinking text. A real failure — llm exiting non-zero — must still die.
+#
+# call_llm is extracted from bin/shellm and exercised directly (sourcing the
+# whole script would run it) against a stub `llm` on PATH.
+
+set -uo pipefail
+
+HERE="$(cd "$(dirname "$0")" && pwd)"
+REPO="$(dirname "$HERE")"
+
+pass=0
+fail=0
+ok()  { pass=$((pass+1)); printf 'ok   %s\n' "$1"; }
+bad() { fail=$((fail+1)); printf 'FAIL %s%s\n' "$1" "${2:+ — $2}"; }
+
+tmp=$(mktemp -d)
+trap 'rm -rf "$tmp"' EXIT
+
+mkdir -p "$tmp/bin"
+cat > "$tmp/bin/llm" <<'STUB'
+#!/usr/bin/env bash
+# Stub llm: LLM_STUB_MODE picks which provider outcome to imitate.
+case "$LLM_STUB_MODE" in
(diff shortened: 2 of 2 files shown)


nickjalbert:
LGTM!  Thanks!

PR #88, opened 2026-08-28
slack: land emoji reactions in the mind log

headlong42:
Subscribe to `reaction_added` so emoji on the bot's messages (and any DM reaction) land in the mind log as a `:emoji:` body on the same `from_name` / thread as the reacted-to message.

- Channel reactions on other people's messages are dropped (same gate as un-mentioned channel traffic).
- Parent-thread lookup runs in the delivery worker so bolt can ack without waiting on `conversations.history`.
- Needs bot scope `reactions:read` — reinstall / re-grant after merge or Socket Mode will not deliver the events.

Pushed from `headlong42/headlong` because `headlong42` has pull-only on `laude-institute/headlong`.


nickjalbert (changes requested):
Thanks @headlong42.  Need some changes.

From review:
>  
> 1. `slack/src/headlong_slack/inbound.py` (`_item_thread_ts`, ~line 192) — replace `conversations.history` with a lookup that actually resolves a reply's parent (`conversations_replies(channel=..., ts=ts, limit=1)` then `messages[0]["thread_ts"]`), or a ts→thread_ts map built from messages this process already saw. `conversations.history` returns only channel-level messages; `outbound.py:83-86` posts channel replies with `thread_ts`, so for reactions on the bot's own replies the lookup returns nothing and `thread_ts = parent or item_ts` (line 258) encodes a reply ts as the thread root — wrong `from_name`, wrong outbound `thread_ts`, and the `(on <item_ts>)` branch never fires.
> 2. `slack/src/headlong_slack/inbound.py` (`_deliver`/`_item_thread_ts`) — bound the lookup: pass an explicit short timeout and fail open to `item_ts`, or resolve off the delivery path. There is exactly one drain thread (lines 148-151), so every channel reaction puts an untimed Web API round trip in front of all queued human messages, and reaction volume is free for any workspace member.
> 3. `slack/src/headlong_slack/inbound.py:272` — do not `threads.touch(...)` for reaction inbounds (or refresh only an already-active key). `state.py:61-67` inserts unconditionally after pruning, so a thumbs-up creates or resurrects an active thread and, under `SLACK_THREAD_FOLLOWUPS=1`, opens the un-mentioned-reply gate at `inbound.py:231-234` on a signal carrying no request intent.
> 4. `slack/tests/test_inbound.py` — add worker-path coverage. Today the file imports only `reaction_text`/`reaction_to_inbound`; `test_outbound.py` already establishes the fake-client pattern. Cover: reaction on a threaded reply (parent resolved, `from_name` and body rewritten), lookup empty/raising (graceful fallback, no exception), DM (no lookup), and a duplicate `reaction_added` envelope dropped by dedupe.
> 5. `slack/src/headlong_slack/inbound.py:159-160` — the docstring's claim that "last-word can treat it as reaction-only, not a request" describes machinery that does not exist: no such component is in the tree, and `thinkers/responder/step` fires on every message step addressed to the identity with a NO_REPLY rule that covers only bare acknowledgments. Either implement the gate (name `:emoji:` bodies as acknowledgment-class in the responder prompt) or delete the claim, so the cost/behavior of every reaction is stated accurately.
>

nickjalbert (changes requested):
requested changes form review:

> 1. **Escape the Markdown backticks in the responder prompt** ([thinkers/responder/step:309](https://github.com/laude-institute/headlong/blob/36a9a409d93a70aaec0fceb871b8078fb8f4126b/thinkers/responder/step#L309))
>
> The new backticks are inside the double quoted `reply_system` shell string, so Bash treats `:emoji:` and `(on <ts>)` as command substitutions. Every responder run now exits while building its prompt, before it calls the model or sends a reply.
>
> Escape both pairs of backticks as the nearby `DEFER` examples do, or remove the Markdown formatting. Verify the fix with `shellcheck -S warning thinkers/responder/step`, `SHELLM_ENV=local tests/test_responder_metrics.sh`, and the full CI suite.
>
> 2. **Bound worker creation after lookup timeouts** ([slack/src/headlong_slack/inbound.py:220](https://github.com/laude-institute/headlong/blob/36a9a409d93a70aaec0fceb871b8078fb8f4126b/slack/src/headlong_slack/inbound.py#L220))
> 
> `shutdown(wait=False)` does not stop a running Slack call. Since each reaction creates a new executor, repeated timeouts leave one live worker for each lookup until the Slack SDK call returns. I reproduced eight live worker threads after eight timed out calls.
>
> A shared executor with a fixed worker count, or an HTTP level timeout, would bound the resource use. A test with a deliberately delayed client would verify the timeout path. The Slack client also has its own 30 second timeout, so this can wait until after the responder failure is fixed.
>

Then please make sure CI is green and we should be good to merge.  thanks!

nickjalbert (changes requested):
few more changes:

  > 1. **Do not queue stale lookups after a timeout** ([slack/src/headlong_slack/inbound.py:228](https://github.com/laude-institute/headlong/blob/f163f93a6da149f79ff06da3aaee5b42b8aa8e5e/slack/src/headlong_slack/inbound.py#L228))
  >
  > The shared executor limits running threads to one, but its pending work queue is unbounded. Each timed out reaction submits another `reactions_get` call while the earlier call is still running.
  >
  > I reproduced this on the current head. Eight 50 ms timeouts against one 800 ms call left one call running and seven stale calls queued. Continued reaction traffic can grow the queue and issue old Slack requests after delivery has already used the fallback. The drain thread also waits up to two seconds for each reaction, so a burst delays later human messages by two seconds per item.
  >
  > Keep at most one lookup future pending and use the fallback immediately while that future is still running, or use another design with a fixed pending work limit. Extend the timeout test so the bridge stays running after the timeouts. Assert that pending work and started Slack calls remain bounded, then enqueue a normal message and verify that it is delivered after at most one short timeout.
  >
  > 2. **Do not run reaction permalink lookups synchronously on the shared worker** ([slack/src/headlong_slack/inbound.py:315](https://github.com/laude-institute/headlong/blob/f163f93a6da149f79ff06da3aaee5b42b8aa8e5e/slack/src/headlong_slack/inbound.py#L315))
  >
  > Every accepted reaction also calls `chat_getPermalink` on the single drain thread. The Slack client has a 30 second default timeout. The call existed before this PR, but the PR lets high volume reaction traffic reach it. During a Slack API problem, each reaction can therefore delay every later human message by another 30 seconds.
  >
  > Keep a reaction marker on `InboundMessage`, then skip the permalink for reactions or resolve it through bounded background work. Add a test with a delayed `chat_getPermalink`, followed by a normal inbound message, and verify that reaction traffic does not hold the normal message in the queue.

nickjalbert:
LGTM!

PR #89, opened 2026-08-29
dash timeline: deeplink to steps and runs

nickjalbert:
The timeline URL now mirrors the open detail modal as ?step= or ?run=, so the address bar is always a shareable link, and the modal has a copy link button. 

Opening a link scrolls the canvas to the element, pulses it, and opens its detail modal. 

A step older than the loaded window falls back to the fetch one step modal that search already uses, and a run older than the window gets a notice to page history in with load older, so a link never drags thousands of old steps into the page.

### DIFF
--- a/web/viewer/app/app.css
+++ b/web/viewer/app/app.css
@@ -287,6 +287,18 @@ html.dark {
   transform-origin: left center;
 }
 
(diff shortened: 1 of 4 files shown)


nickjalbert:
LGTM!

PR #90, opened 2026-08-29
fast-follows from the #89 review

RevEngineBot:
Fast-follows from the committee review of #89, applied after its merge.

## The plan implemented

1. **Do not consume the pending deeplink against an empty layout** — `web/viewer/app/components/timeline-view.tsx`. Bail out (leaving the ref set) while the layout has nothing to match against, so the resolver retries when cells/blocks arrive. Verify: mount the timeline with `?step=<id-in-window>` while the mindlog query is still in flight; the step modal must open on the real step, not the "older than the loaded window" modal.

```diff
   useEffect(() => {
     const want = pendingDeeplink.current;
     if (!want) return;
+    if (!want.step && !want.run) {
+      pendingDeeplink.current = null;
+      return;
+    }
+    // Nothing to match against yet: keep the deeplink pending so a
+    // still-loading first layout doesn't produce a false "missing".
+    if (layout.cells.length === 0 && layout.blocks.length === 0) return;
     pendingDeeplink.current = null;
-    if (!want.step && !want.run) return;
```

2. **Reconcile selection from the query params, not just at mount** — same file. Add an effect keyed on `[stepParam, runParam]` that re-runs the same resolve path (or clears `selected`/`missing` when both are null), guarded so it does not re-apply an already-selected item and does not fight `applySelection`'s own writes. Verify: with the timeline mounted, change `?step=` to a different in-window id via a soft navigation and via Back/Forward — the modal must follow in both directions, and clearing the params must close it. Note for the implementer: nuqs defaults to `history: 'replace'`, so today's open/close does not push entries; the observable break is soft navigation and landing on a history entry that already carried a param.

3. **Add viewer tests for the new behaviors** — `web/viewer` (locate or state the absence of a harness). Cover: `?step=` for a step in the loaded window (modal opens, row scrolled, flash class applied); `?step=` for a machinery step with no cell (lands on its block); `?run=` for a loaded run; unknown step id → `StepModal` fallback; unknown run id → "load older" notice; `pinned === false` on arrival with either param and `true` without; open/close writes and clears the params. Verify: tests fail against the pre-fix code from item 1 and pass after.

4. **Never let a missing-step deeplink render nothing** — `timeline-view.tsx`, the `missing?.kind === "step" && traj` branch. When `traj` is null, render the run-branch style notice instead of nothing. Verify: with no traj context, `?step=<unknown>` shows an explanatory modal rather than a blank screen with a stale param.

5. **Normalize the URL when the resolver selects** — `timeline-view.tsx`. Use `applySelection(...)` in both resolver branches (or explicitly clear the losing param) so a URL carrying both `?step=` and `?run=` ends up with exactly one, keeping the copy-link contract honest. Optionally switch to nuqs `useQueryStates({ step, run })` so one selection is one URL write. Verify: load `?step=A&run=B`; after resolution the address bar holds only `?step=A`.

6. **Handle clipboard rejection** — `web/viewer/app/components/timeline-detail.tsx`, `CopyLinkButton`.

```diff
-        void navigator.clipboard?.writeText(window.location.href).then(() => {
-          setCopied(true);
-          setTimeout(() => setCopied(false), 1500);
-        });
+        void navigator.clipboard
+          ?.writeText(window.location.href)
+          .then(() => {
+            setCopied(true);
+            setTimeout(() => setCopied(false), 1500);
+          })
+          .catch(() => {});
```
Verify: deny clipboard permission (or blur the document) and click — no unhandled rejection in the console.

_Opened by rev-engine (test suite green before push: `SHELLM_ENV=local bash tests/run-all.sh`). This PR goes through the same review pipeline as any other change._

### DIFF
--- a/web/viewer/app/components/timeline-detail.tsx
+++ b/web/viewer/app/components/timeline-detail.tsx
@@ -59,10 +59,13 @@ function CopyLinkButton() {
       title="Copy link to this item"
       className="absolute right-10 top-3 rounded-md p-1 text-muted-foreground hover:bg-accent hover:text-foreground"
       onClick={() => {
-        void navigator.clipboard?.writeText(window.location.href).then(() => {
-          setCopied(true);
-          setTimeout(() => setCopied(false), 1500);
-        });
+        void navigator.clipboard
+          ?.writeText(window.location.href)
+          .then(() => {
+            setCopied(true);
+            setTimeout(() => setCopied(false), 1500);
+          })
+          .catch(() => {});
       }}
     >
       {copied ? (
--- a/web/viewer/app/components/timeline-view.test.tsx
+++ b/web/viewer/app/components/timeline-view.test.tsx
@@ -1,0 +1,293 @@
+// Deeplink behavior of the Timeline tab: ?step=/?run= resolution at mount
+// (including a first layout that hasn't loaded yet), reconciliation with
+// later URL changes (soft navigation, Back/Forward), fallbacks for targets
+// outside the loaded window, follow-mode pinning, and the URL round-trip
+// of opening/closing the detail modal.
+
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { cleanup, render, screen, waitFor } from "@testing-library/react";
+import { fireEvent } from "@testing-library/dom";
+import { NuqsTestingAdapter } from "nuqs/adapters/testing";
+import { MemoryRouter } from "react-router";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+import { TimelineView } from "~/components/timeline-view";
+import { buildTimeline, type TimelineLayout } from "~/lib/timeline-model";
+import { TrajContext, type TrajContextValue } from "~/lib/traj-context";
+import type { NormalizedStep, RunGroup, StepSource, StepType } from "~/lib/types";
+
+// jsdom ships no ResizeObserver; ExpandableText measures with one.
+globalThis.ResizeObserver ??= class {
+  observe() {}
+  unobserve() {}
+  disconnect() {}
+} as unknown as typeof ResizeObserver;
+
+// The fetch-one fallback (StepModal) and run-command hydration hit the
+// network; fail the former so the fallback's terminal state is stable.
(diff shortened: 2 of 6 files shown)


nickjalbert:
From review:

> The deeplink reconciliation is sound on the normal path: `appliedParams` is written synchronously in `applySelection` before the nuqs setters, so the reconcile effect does not re-trigger on the component's own URL echoes, and the "step wins, losing param cleared" normalization is applied consistently in both the resolver and the missing-target branches. The new test file is a genuine regression guard — the empty-layout hold, URL-follow, both-params normalization, and clipboard-rejection tests all exercise paths that fail or don't exist pre-fix.
>
> Three things remain, all traced to one gate:
>
> - The empty-layout hold (`timeline-view.tsx:439`) never releases for a trajectory that is loaded and genuinely empty — `routes/timeline.tsx:63-68/158` mounts `TimelineView` only after loading, so an empty layout means empty data, not pending data. A `?step=`/`?run=` on such a trajectory now renders nothing at all (pre-fix it showed the fetch-one fallback). Contributor-side or maintainer-side: gate on a real loading signal.
> - While the deeplink is held, the reconcile effect bails (`if (pendingDeeplink.current) return`), so params that change or clear during the hold are ignored; on a clear, the mount pass later resolves the stale target and `applySelection`'s queued `setStepParam` is never cancelled by the reconcile clear branch, so the stale param is written back. Same fix as above. Contributor-side or maintainer-side.
> - A mount URL carrying both params resolves twice in one commit (duplicate scroll + flash restart, identical end state, verified idempotent). Cosmetic; collapsing `step`/`run` into `useQueryStates` removes it. Maintainer-side.
> - `bun run test` is not run anywhere automated: the viewer CI job runs typecheck + build only, and `tests/run-all.sh` globs `test_*.sh`, so the cited green suite says nothing about this file. Maintainer-side (needs workflow scope).
>
> Two round-1 concerns were checked and dropped: `hasMemory` is a real prop on the pinned nuqs 2.10.1 testing adapter, and the pending-deeplink test's synchronous `"loading…"` assertion *is* load-bearing pre-fix.

nickjalbert:
LGTM!

PR #91, opened 2026-08-29
dash timeline: copy links to run substeps

nickjalbert:
Steps inside a run modal now show a copy link button on hover. The link opens the step in place when it is loaded, or fetches it alone when it is older than the loaded window.


nickjalbert:
LGTM!

PR #92, opened 2026-08-29
fast-follows from the #91 review

RevEngineBot:
Fast-follows from the committee review of #91, applied after its merge.

## The plan implemented

1. **Deduplicate copy-link logic** (`web/viewer/app/components/timeline-detail.tsx`): extract the shared "write to clipboard, flash `copied` state for 1500ms" logic out of `CopyLinkButton` and `SubstepLink` into a small shared hook (e.g. `useCopyLink(getUrl: () => string)`) or a shared `<CopyButton>` component parameterized by URL-building and styling. Verify by confirming both call sites shrink to passing a URL-builder/className and behavior (icon swap, timeout reset) is unchanged. Contributor-side or maintainer-side, whichever picks it up first.
2. **Add a test for the substep deeplink URL construction** (`web/viewer` test suite, if one exists — check for Vitest/RTL config): assert that clicking the substep copy button produces a URL with `step=<step_id>` set and `run` absent, and does not disturb other unrelated query params. If no frontend test harness exists in this repo yet, note that as a prerequisite rather than skipping silently. Maintainer-side to scope (harness existence is unclear from this diff alone).

_Opened by rev-engine (test suite green before push: `SHELLM_ENV=local bash tests/run-all.sh`). This PR goes through the same review pipeline as any other change._

### DIFF
--- a/web/viewer/app/components/timeline-detail.tsx
+++ b/web/viewer/app/components/timeline-detail.tsx
@@ -48,25 +48,33 @@ export type TimelineSelection =
   | { kind: "step"; step: NormalizedStep }
   | { kind: "run"; block: TimelineBlock };
 
+/** Copy `getUrl()` to the clipboard and flash `copied` for 1.5s; a denied
+ * clipboard (permissions, non-secure context) is silently ignored. */
+function useCopyLink(getUrl: () => string) {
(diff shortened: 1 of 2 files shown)


nickjalbert:
LGTM!

PR #93, opened 2026-08-29
Add documentation for local models

danielhstahl:
Added ollama and llama.cpp details


nickjalbert:
Thanks!  Pushed a diff to reflect updated install flow and flesh out some commands.

PR #94, opened 2026-08-29
init: first-class local (OpenAI-compatible) model server setup

### LINKED ISSUE #65: Add first-class Ollama provider support
Headlong's persistent-agent model is a particularly good fit for local inference, but `bin/llm` currently supports Anthropic, OpenAI, Gemini, and OpenRouter only. Ollama exposes an OpenAI-compatible `/v1/chat/completions` API, so adding it as a first-class provider should be relatively small and would avoid requiring users to masquerade Ollama as OpenAI with a dummy API key.

## Proposed configuration

```bash
SHELLM_PROVIDER=ollama
SHELLM_MODEL=qwen3:8b
OLLAMA_HOST=http://localhost:11434
```

Remote hosts should work as well:

```bash
OLLAMA_HOST=http://madhatter.local:11434
```

Provider selection should prefer:

1. explicit `--provider`
2. `SHELLM_PROVIDER`
3. model-name inference
4. the existing API-key-based fallback

An explicit provider is important because Ollama model names such as `qwen3:8b`, `gemma3:12b`, and `llama3.2` do not encode the provider in a way `detect_provider()` can reliably infer.

## Implementation sketch

Because Ollama's chat endpoint is OpenAI-compatible, most of the existing OpenAI path can be reused:

```bash
build_payload_ollama()   { build_payload_openai; }
extract_text_ollama()    { extract_text_openai; }
check_response_ollama()  { check_response_openai; }
check_truncated_ollama() { check_truncated_openai; }
stream_ollama()          { stream_openai; }
```

`get_url_headers()` would need an Ollama branch approximately like:

```bash
ollama)
    url="${LLM_API_URL:-${OLLAMA_HOST:-http://localhost:11434}/v1/chat/completions}"
    headers=(-H "Content-Type: application/json")
    ;;
```

No API key should be required.

The existing OpenAI-compatible streaming behavior, retry protections, newline preservation, and usage handling should remain in effect. Where Ollama returns OpenAI-style usage data, it should continue into Headlong's usage ledger with `provider: "ollama"` and the actual local model name.

Unknown Ollama models should not silently inherit an unnecessarily small output cap if an explicit limit is configured; the existing `LLM_MAX_TOKENS` override may be sufficient initially.

## Installer / setup

`headlong-init` should offer Ollama as a provider and, when selected:

- skip the API-key prompt
- ask for `OLLAMA_HOST` (default `http://localhost:11434`)
- ask for the model name
- verify the host/model is reachable
- persist `SHELLM_PROVIDER`, `SHELLM_MODEL`, and `OLLAMA_HOST`

Headlong should not attempt to install Ollama or pull models automatically.

## Tests

Mocked tests should cover:

- `--provider ollama`
- `SHELLM_PROVIDER=ollama`
- arbitrary Ollama model names
- operation with no cloud API keys set
- default and remote `OLLAMA_HOST`
- streaming and non-streaming output
- newline preservation / `[DONE]`
- empty-response and retry behavior
- usage accounting
- unchanged behavior for the existing providers

## Acceptance criterion

With an Ollama server already running, this should work without any cloud API key:

```bash
SHELLM_PROVIDER=ollama \
SHELLM_MODEL=qwen3:8b \
OLLAMA_HOST=http://localhost:11434 \
shellm "hello"
```

The persistent identity should then be able to use the same configuration normally.

This issue is intentionally limited to first-class Ollama support. Tiered routing between local background cognition and frontier-model escalation would be a separate architectural proposal.

ghandolphAI:
## What

Installer/setup support for local OpenAI-compatible model servers (Ollama, LM Studio, vLLM, llama.cpp, proxies). The runtime half of #65 (the generic `openai-compatible` provider) landed in #78; this adds the missing half — `headlong-init` only knew the cloud API-key flow, so a local setup had to be configured by hand after install.

## How it works

- **Explicit choice, never inferred** (per design/providers.md): a `Use a local model server? [Y/n]` prompt, `HEADLONG_PROVIDER=local|cloud`, or keep-when-healthy on re-runs.
- **Local path**: asks for the server's address as the operator knows it (`http://localhost:1234`, a LAN address, or a full URL — `/v1` and `/chat/completions` are added automatically if missing), an optional bearer key (hidden input), then **verifies by fetching `GET /v1/models`** and lets the operator **pick the model from that list** (numbered menu, free-text fallback for huge catalogs). Servers that hide `/v1/models` fall back to a typed/pinned model.
- **Sandbox-aware URL**: probes run host-side through a URL's `localhost` twin (`host.docker.internal` only resolves inside containers); with the Docker sandbox on, the persisted `SHELLM_API_URL` is rewritten `localhost`/`127.0.0.1` → `host.docker.internal` (announced), so the agent dials an address that resolves. LAN addresses and explicit container URLs pass through untouched.
- **Persisted**: `LLM_PROVIDER=openai-compatible`, `SHELLM_API_URL`, `LLM_API_KEY` (only when given — a stale key is removed so the endpoint is keyless, not mis-keyed), `SHELLM_MODEL` — then proven with a real chat completion through `bin/llm`, the same handshake a thinker makes.
- **Paths**: host installs, the long-lived Docker container, and no-tty/CI installs (`HEADLONG_PROVIDER=local` + `HEADLONG_LOCAL_URL` is all a scripted install needs). No-tty: an unreachable server is a hard stop; a reachable server without a model list only warns.
- **Install-location**: `install.sh --init` from a checkout now offers the same container / this-machine / no-sandbox menu the one-liner makes, and the container path defaults the forwarded `HEADLONG_REPO`/`HEADLONG_BRANCH` to the checkout's own origin and branch — so choice 1 installs and runs **this** code, not upstream main (an operator-exported `HEADLONG_REPO` still wins).
- **Idempotent**: a re-run keeps a working local config without re-asking; re-running with `HEADLONG_PROVIDER=local` repairs a broken one.

Headlong still never installs a server or pulls models (per the issue).

## Infrastructure consistency

The installer writes exactly the variables the existing infrastructure consumes — no new runtime names: `LLM_PROVIDER`/`SHELLM_API_URL`/`SHELLM_MODEL`/optional `LLM_API_KEY`, the tuple documented in `design/providers.md` and `docs/shellm.md` (`LLM_API_URL` stays a direct-`llm` concern and is deliberately not written). Verified by execution that `bin/llm`, `bin/shellm` (incl. the `SHELLM_API_URL`→`LLM_API_URL` re-export and sandbox/nested-run forwarding), `thinkers/_lib/common.sh`, and the dash's `llm_probe` all consume this tuple with no changes. `HEADLONG_PROVIDER`/`HEADLONG_LOCAL_URL`/`HEADLONG_LOCAL_MODEL`/`HEADLONG_LOCAL_API_KEY` are installer inputs only, like `HEADLONG_IDENTITY_*`.

## Tests

New `tests/test_local_llm_setup.sh` (34 checks, curl/llm/identity/headlong-web stubbed, no network): happy path + persistence, cloud-key non-leak, dead-endpoint hard stop, URL normalization (bare address, trailing slash, full URL), sandbox rewrite to host.docker.internal, explicit container URL passthrough, local key persistence, no-pin → first listed model, failing-completion detection, explicit cloud override, re-run idempotence. All pass; shellcheck clean.

Full `tests/run-all.sh`: all suites pass except `test_thinkers_pending.sh` ("queue-full drop logged"), which also fails on unmodified main in this environment — pre-existing timing flake, unrelated.

Proven end-to-end on a real install (Docker-container path, LM Studio server on the LAN): location menu → container cloned this branch → local offer → bare `http://<lan>:1234` accepted → 11-model picker from the live server → verified completion → identity created → mind running with no cloud key; `ada say` round-trips through the local model.

Docs: `docs/install.md` documents both interactive and non-interactive (env-var) local flows.

Closes #65 (installer half; runtime completed by #78).

### DIFF
--- a/bin/shellm
+++ b/bin/shellm
@@ -109,11 +109,57 @@ die() { echo "shellm: error: $*" >&2; exit 1; }
 
 # A local model bound to the Docker host is usually configured as localhost
 # for thinkers and other llm calls that run on that host. Code executed in a
-# shellm sandbox needs Docker's hostname for the same server.
+# shellm sandbox needs Docker's hostname for the same server. The rewrite
+# touches ONLY a loopback authority host — a substring replace would mangle
+# hosts like mylocalhost.dev or path segments containing the name, and
+# redirect a Bearer-keyed request to a host nobody configured. The same
+# helpers exist in tools/headlong-init (this repo duplicates small helpers
+# rather than sharing a lib; keep the copies identical).
+
+# _shellm_url_host <url> — the authority's host, IPv6 brackets stripped.
+_shellm_url_host() {
+    local rest="${1#*://}"
+    local authority="${rest%%[/?#]*}"
+    local hostport="${authority##*@}"
+    case "$hostport" in
+        \[*\]*) hostport="${hostport#\[}"; printf '%s' "${hostport%%]*}" ;;
+        *)      printf '%s' "${hostport%%:*}" ;;
+    esac
+}
+
+# _shellm_url_host_swap <url> <newhost> — replace only the authority's host.
+_shellm_url_host_swap() {
+    local url="$1" newhost="$2"
+    local rest="${url#*://}" scheme=""
+    [[ "$rest" == "$url" ]] || scheme="${url%%://*}://"
+    local authority="${rest%%[/?#]*}"
+    local tail="${rest#"$authority"}"
+    local userinfo=""
+    case "$authority" in
+        *@*) userinfo="${authority%@*}@" ;;
+    esac
+    local hostport="${authority##*@}" port=""
+    case "$hostport" in
+        \[*\]:*) port=":${hostport##*:}" ;;
+        \[*\])   port="" ;;
+        *:*)     port=":${hostport##*:}" ;;
+    esac
+    printf '%s%s%s%s%s' "$scheme" "$userinfo" "$newhost" "$port" "$tail"
+}
+
+# _shellm_url_is_loopback <url> — does the host name this machine's loopback?
+_shellm_url_is_loopback() {
+    case "$(_shellm_url_host "$1")" in
+        localhost|127.*|0.0.0.0|::1) return 0 ;;
+    esac
+    return 1
+}
+
 _shellm_container_api_url() {
     local url="$1"
-    url="${url//localhost/host.docker.internal}"
-    url="${url//127.0.0.1/host.docker.internal}"
+    if _shellm_url_is_loopback "$url"; then
+        url=$(_shellm_url_host_swap "$url" host.docker.internal)
+    fi
     printf '%s' "$url"
 }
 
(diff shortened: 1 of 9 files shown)


nickjalbert:
LGTM! Thanks for the PR. I pushed some hardening on top (quoting the .env writes, host-only URL rewrite, and validating before persisting) and will merge once green.

PR #96, opened 2026-08-31
Fix missing thinker tools and CHATRC in Docker runs

shaircast:
## Summary

- forward the identity-specific `CHATRC` into generated-code runs
- stage the advertised `focus` and `recap` tools in thinker Docker environments
- add a coupling test covering the stock tool set and chat configuration

## Problem

`identity` exports `<identity>/chat/.chatrc`, and `chat` reads that path through `CHATRC`, but `_build_shellm_flags` did not forward the variable. Docker-backed thinker runs therefore fell back to `./.chatrc` and could not resolve `default_send_from`.

The README and stock identity prompt also advertise `focus` and `recap`, but neither executable was included in the thinker `--bin` list. They were available in local mode and silently absent in Docker mode.

## Testing

- `bash tests/test_thinker_tool_coupling.sh` — 14 passed
- `bash tests/test_thinker_env_fallback.sh` — 8 passed
- `bash tests/test_monolith_backoff.sh` — 10 passed
- clean-environment `bash tests/run-all.sh` — 40 suites passed, 0 failed

### DIFF
--- a/tests/test_thinker_tool_coupling.sh
+++ b/tests/test_thinker_tool_coupling.sh
@@ -1,0 +1,74 @@
+#!/usr/bin/env bash
+# tests/test_thinker_tool_coupling.sh — every stock tool promised to the mind
+# must be forwarded into shellm's Docker sandbox.
+#
+# Why: _build_shellm_flags constructs the repeated --bin flags used by every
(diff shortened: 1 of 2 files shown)


nickjalbert:
LGTM!  Thanks!

PR #98, opened 2026-08-31
Chat composer: Shift+Enter for newline, Enter to send

sandeep:
## What

Both chat composers (the desktop `/chat` page and the `/talk` PWA) used a
single-line `<Input>`, so there was no way to type a multi-line message at
all. This swaps in an auto-growing `<Textarea>`: Enter still sends, and
Shift+Enter inserts a newline.

## How it works

- New `Textarea` UI primitive (`components/ui/textarea.tsx`), matching the
  existing shadcn-style `Input`/`Button` components in this dir.
- New `useAutosizeTextarea` hook (`hooks/use-autosize-textarea.ts`) that grows
  the box to fit content via `scrollHeight`, capped by `max-h-40` (then
  scrolls internally). Deliberately not CSS `field-sizing: content` — that's
  unsupported on older Safari/iOS, and `/talk` is explicitly a phone PWA.
- Both composers keep their existing submit logic; the textarea's `onKeyDown`
  calls `form.requestSubmit()` on Enter (no Shift) instead of duplicating the
  trim/pending guard.

## Tests

New `app/routes/chat.test.tsx` (2 checks): Enter sends and clears the draft;
Shift+Enter does not send and leaves the draft intact. `bun run typecheck`,
`bun run build`, and the full `vitest` suite (15 tests) all pass.

### DIFF
--- a/web/viewer/app/components/ui/textarea.tsx
+++ b/web/viewer/app/components/ui/textarea.tsx
@@ -1,0 +1,18 @@
+import * as React from "react"
+
+import { cn } from "~/lib/utils"
+
+function Textarea({ className, ...props }: React.ComponentProps<"textarea">) {
+  return (
+    <textarea
+      data-slot="textarea"
+      className={cn(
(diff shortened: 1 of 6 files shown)


nickjalbert:
LGTM!  Added fix to preserve expected behavior on mobile/PWA.  Thanks!

PR #100, opened 2026-09-01
llm: fix Anthropic thinking/effort payload to match the real API

### LINKED ISSUE #99: shellm sends invalid thinking and effort fields to the Anthropic API
## Description

Monolith wakeup fails when THINK_MODEL or SHELLM_MODEL is set to a real
Anthropic Claude model. bin/llm sends two fields the Anthropic API
rejects.

## Steps to Reproduce

1. Set THINK_MODEL (or SHELLM_MODEL) to a real Anthropic Claude model,
   e.g. claude-sonnet-4-5.
2. Start thinkers.
3. Watch monolith's log.

## Expected Behavior

monolith completes a normal shellm run.

## Actual Behavior

Monolith wakeup fails with run FAILED (rc=1). The dispatcher backs off
and retries. It fails again.

bin/shellm calls bin/llm with --thinking --effort high
(bin/shellm:1375). bin/llm sends:

1. thinking:{type:"adaptive"} → API error: "adaptive thinking is not
   supported on this model"
2. output_config:{effort:"high"} → API error: "This model does not
   support the effort parameter"

Neither field exists in the Anthropic Messages API.

## Environment

- Model: claude-sonnet-4-5 (Anthropic)
- Affected code: bin/llm, bin/shellm, thinkers/monolith/step
- Verified live on a claude-sonnet-4-5 identity

## Root Cause

supports_anthropic_thinking() in bin/llm checks the model name. It does
not check the payload fields.

type:"adaptive" and output_config.effort came in with commit 92c8c73
("adapted from box commit 1e3c68b"), merged 2026-08-17 — 15 days ago.

CI stubs curl and llm. No test calls the real API.

## Additional Context

responder does not use this code path. responder calls bin/llm directly
(thinkers/responder/step). It does not pass --thinking or --effort. Chat
replies keep working.

monolith (the mind) runs through bin/shellm (thinkers/monolith/step:316).

## Proposed Fix

A PR is coming.

- Change thinking.type to "enabled". Add a required budget_tokens field.
- Add a new guard, supports_anthropic_effort(). It returns false for
  every current model.
- Gate output_config.effort behind this guard, like the existing
  thinking guards.

sandeep:
Fixes #99

## What

`thinking:{type:"adaptive"}` and `output_config:{effort:...}` are not real
fields in Anthropic's public Messages API. The real API rejects both on
claude-sonnet-4-5, its flagship model. `shellm` always calls `bin/llm`
with `--thinking --effort high`, so monolith wakeup fails on any real
Anthropic Claude model. `responder` calls `bin/llm` directly, without
`--thinking`/`--effort`, so chat replies keep working and can look like
everything is fine.

## How it works

- `thinking.type` is now `"enabled"`, with a required `budget_tokens` —
  half of `max_tokens`, floored at the API's 1024 minimum, leaving the
  other half of the token budget for the visible response.
- `output_config.effort` is now gated behind a new
  `supports_anthropic_effort()` guard, mirroring the existing
  `supports_anthropic_thinking()`/`supports_openai_reasoning()`/
  `supports_gemini_thinking()` guards. It returns false for every known
  model, since none of them accept this field via the real API.
  `LLM_ASSUME_THINKING=1` is the same escape hatch already used for the
  other guards, for a future model that does support it.
- Help text (`--thinking`, the Anthropic section under "Thinking &
  effort") updated to match.

## Tests

New `tests/test_llm_anthropic_thinking.sh` (9 checks, curl stubbed to
capture the outgoing payload): `thinking.type` is `"enabled"` not
`"adaptive"`, `budget_tokens` is half of `max_tokens` and floors at 1024,
`--effort` no longer adds `output_config` and is reported on stderr,
`thinking` still sends when `--effort` is dropped alongside it (the exact
combination `shellm` always sends), and `LLM_ASSUME_THINKING=1` forces
`output_config.effort` through.

Verified against the real Anthropic API (not just the stub): both flags
together on `claude-sonnet-4-5` now succeed end to end, live on a
previously-broken identity — monolith went from 100% `rc=1` to a normal
multi-iteration run immediately after.

Full `tests/run-all.sh`: 37/38 pass; the one failure
(`test_thinkers_pending.sh`) is unrelated dispatcher-queue timing, and
fails the same way on unmodified `main` in this environment.
`shellcheck -S warning`, `bash -n`, and a bash-3.2 parse check are all
clean.

### DIFF
--- a/bin/llm
+++ b/bin/llm
@@ -156,7 +156,7 @@ Options:
   --provider NAME          Provider: anthropic, openai, gemini, openrouter,
                            opencode, opencode-go, openai-compatible, or
                            adapter (needs LLM_ADAPTER)
-  --thinking [LEVEL]       Enable thinking (Anthropic: adaptive; Gemini: minimal/low/medium/high)
+  --thinking [LEVEL]       Enable thinking (Anthropic: on/off; Gemini: minimal/low/medium/high)
   --effort LEVEL           Thinking effort/provider output effort
   --raw                    Print raw non-streaming API response JSON
   -h, --help               Show this help
@@ -180,12 +180,13 @@ Thinking & effort:
   These flags control how much reasoning the model does before answering.
   Each provider implements this differently:
 
-  Anthropic (--thinking, --effort):
-    --thinking enables extended thinking (chain-of-thought visible on stderr).
-    --effort controls output effort independent of thinking:
-      low, medium, high (default), xhigh, max
-    Both flags can be combined. Without --thinking, the model still reasons
-    internally but doesn't expose it.
+  Anthropic (--thinking):
+    --thinking enables extended thinking (chain-of-thought visible on
+    stderr), budgeted at half of --max-tokens. Without --thinking, the model
+    still reasons internally but doesn't expose it. --effort is not a real
+    Anthropic API field (the API rejects it on every current model) and is
+    always ignored here; LLM_ASSUME_THINKING=1 forces it through anyway if a
+    future model ever supports it.
 
   Gemini (--thinking LEVEL):
     Gemini models think by default. --thinking sets the depth:
@@ -474,6 +475,16 @@ supports_anthropic_thinking() {
     esac
 }
 
+# output_config.effort is not a documented field in Anthropic's public
+# Messages API (verified against the real API on claude-sonnet-4-5, its
+# flagship model: "This model does not support the effort parameter" — the
+# same kind of ported assumption as the old thinking type:"adaptive" below).
+# No known model supports it, so this always returns 1 unless overridden.
+supports_anthropic_effort() {
+    [[ "${LLM_ASSUME_THINKING:-0}" == 1 ]] && return 0
+    return 1
+}
+
 supports_openai_reasoning() {
     [[ "${LLM_ASSUME_THINKING:-0}" == 1 ]] && return 0
     case "$LLM_MODEL" in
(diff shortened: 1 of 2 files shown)


nickjalbert (changes requested):
Thanks for the PR! From review:

> 
> **The patch breaks Claude Opus 4.7 and new Anthropic models**
> 
> bin/llm:467 treats every current Claude family alike. bin/llm:521 then sends thinking.type: "enabled" with budget_tokens, while bin/llm:483 drops output_config.effort for every model.
> 
> That payload fixes Sonnet 4.5, but Claude 4.7 and later reject manual extended thinking and require adaptive thinking with effort. Headlong defaults to Opus 4.7 and always passes both flags, so normal shellm calls would fail. Anthropic also documents effort support for several current models. Anthropic's API primer describes the model split.
> 
> Add an explicit model capability table. At minimum, test Sonnet 4.5 with enabled plus a budget, and Opus 4.7 with adaptive plus effort. Include the supported 4.6 behavior as well.
> 
> **Small max_tokens values produce an invalid budget.**
> 
> bin/llm:527 floors the thinking budget at 1024 without ensuring it remains below max_tokens. The new test therefore accepts max_tokens: 100 with budget_tokens: 1024.
> 
> Anthropic requires budget_tokens to be at least 1024 and less than max_tokens. Anthropic Messages API reference.
> 
> Reject this combination, disable thinking with a warning, or raise the effective output limit. Add boundary tests for 1024 and 1025 tokens.
> 

After those are addressed and you rebase on main, we should be good to 🚢 !

PR #101, opened 2026-09-01
llm/shellm: add OpenAI Responses completion protocol

24601:
## What

Adds an opt-in OpenAI Responses completion protocol to `bin/llm` and process-local Responses continuation to `shellm`. Chat Completions remains the default; its retry behavior is unchanged.

Set `LLM_API_FORMAT=responses` for llm or `SHELLM_API_FORMAT=responses` for shellm to preserve typed reasoning and response state across turns.

## Completion contract

- Native OpenAI and OpenRouter select their Responses endpoints; `openai-compatible` uses its configured endpoint.
- Typed input is preserved, including multimodal messages, reasoning items, function calls and function outputs.
- `LLM_RESPONSES_BODY_FILE` accepts tools, include, store, text format, metadata, truncation, service tier and compatible extensions. CLI-owned fields stay deterministic; conversation state and background mode are rejected.
- Requests augment include with `reasoning.encrypted_content` for exact stateless replay. OpenRouter routing/privacy policy overlays remain effective.
- Buffered and SSE paths handle text, refusals, reasoning summaries, function-only output, incomplete/failed status, errors and usage.
- `LLM_RESPONSE_FILE` atomically records the terminal Response or error as mode 0600; an unwritable sidecar fails before dispatch or fails the call rather than reporting artifact-free success.

## Adversarial hardening

Responses CREATE is **never automatically retried**, even before output. Silence does not prove the provider did not accept work. Lost acknowledgements, missing terminals and malformed success envelopes fail with `error.code=outcome_unknown`, preserving any known response ID. No provider idempotency guarantee is invented.

Successful buffered/streamed completion requires a nonempty ID, output array, completed/incomplete status and no embedded error. Stream identity and terminal event/status must agree. Cancelled and failed terminals fail without recreation; incomplete remains partial output with a warning. Function-only output remains a valid protocol response.

Shellm retains process-local previous-response state and exact typed replay. A structured pre-generation continuation rejection can trigger one replay fallback, after which the run stays stateless. Unknown outcomes, known IDs, terminal failures and calls that emitted text cannot enter fallback merely because an error mentions a previous response. OpenRouter begins in replay mode; response IDs are not persisted into the durable trajectory.

## Scope boundary

Retrieve, cancel, delete, input-item listing, Conversations, background jobs, compaction and WebSocket transport **remain out of scope** (see PR on my fork for these capabilities I am happy to fold into here, I wanted to keep this in spirit of what the core impl looked like with completions and not raise the LoC limit, either). No SDK/Python dependency, lifecycle CLI or line-budget increase is introduced by this PR.

## Verification

- `test_llm_responses.sh`: **35 passed**.
- `test_llm_responses_faults.sh`: **30 passed**, including lost acknowledgements, retained recovery handles, malformed/unknown terminals, cancellation and stream identity mismatch; exactly one create asserted.
- `test_shellm_responses_continuation.sh`: **20 passed**.
- Full Linux harness with jq 1.7.1: **65 suites passed** in a linked worktree; the remaining existing runtime-prompt test assumes `.git/HEAD` is a file path and fails in linked worktrees. It passed **13/13** unchanged in a normal clone of the same commit.
- Warning-level ShellCheck passes; core remains below **11K** (about **10.7K**).
- [Full CI passed on this head](https://github.com/laude-institute/headlong/actions/runs/34180315402), including Linux/macOS Bash 3.2, Python projects, viewer, TUI, installation, lint and secret scans.

Design, ownership and failure contracts are in `design/responses-api.md`. Existing tests also cover chat retry preservation, OpenRouter policy, shrinking rendered context, uncapped sent-row IDs, preflight sidecar checks, body-file sandbox mounting and stop-after-code-block behavior.

### DIFF
--- a/.agents/skills/operating-headlong
+++ b/.agents/skills/operating-headlong
@@ -1,0 +1,1 @@
+../../skills/operating-headlong
--- a/.agents/skills/operating-headlong-responses
+++ b/.agents/skills/operating-headlong-responses
@@ -1,0 +1,1 @@
+../../skills/operating-headlong-responses
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,6 +72,9 @@ jobs:
         # image; uuidgen (uuid-runtime) is installed explicitly so a nested
         # shellm run never depends on the image's package set.
         run: sudo apt-get install -y -qq uuid-runtime >/dev/null
+      - uses: astral-sh/setup-uv@v7
+        # tools/responses-ws is a uv script (websockets); its test skips
+        # without uv, which would leave the WebSocket transport unexercised.
       - name: Run tests/test_*.sh
         run: tests/run-all.sh
 
@@ -171,6 +174,7 @@ jobs:
         run: |
           brew list coreutils >/dev/null 2>&1 || brew install -q coreutils
           brew list jq >/dev/null 2>&1 || brew install -q jq
+          brew list uv >/dev/null 2>&1 || brew install -q uv
       - name: Make Apple's /bin/bash the `bash` on PATH
         # The image also has Homebrew bash 5 on PATH, which would hide every
         # 3.2 problem. A shim dir ahead of everything else makes `bash` and
@@ -274,7 +278,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 5
     env:
-      LOC_LIMIT: 11500
+      LOC_LIMIT: 12000
     steps:
       - uses: actions/checkout@v5
       - name: Install cloc
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,8 +73,8 @@ jobs:
         # shellm run never depends on the image's package set.
         run: sudo apt-get install -y -qq uuid-runtime >/dev/null
       - uses: astral-sh/setup-uv@v7
-        # tools/responses-ws is a uv script (websockets); its test skips
-        # without uv, which would leave the WebSocket transport unexercised.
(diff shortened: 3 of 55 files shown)


24601:
FYI - I have strived to keep this technically & consistent-with-spirit of the design docs discussing OpenAI compatible API as the core and the providers beyond that doc...since this is kind of a grey area of judgement (e.g. is OpenAI responses a separate provider or falling under the aegis of OpenAI compatible) I've taken an approach that implements the Responses API in the core OpenAI Compatible layer so it can "do the things that it feels like this core portion needs to do" and not much more. 

So, if:

1. This actually is a separate provider in your guys' view (I would hope not and chose not to view it that way for the purposes of raising this PR as we all know where the completions API is destined, hahaha, but there are material, direct functionality, capability, efficiency, and cost upsides in many core functionality headlong use cases and internals that Responses provides, and I've tried to focus on what those are than a kitchen-sink Responses impl, which is a segue to two)

or

2. There are things I've kept out of scope that could be suitable for upstream headlong inclusion, I am doing something larger-scope in my own fork to cover use cases I have that some of the persistent agency work I am doing benefit from directly from the additional things in Responses (capability, efficiency, cost wise): https://github.com/24601/headlong/pull/1

I would love to know/hear.

24601:
Let me know if Responses API ever becomes something you guys want, happy to reopen/revisit, but going to close for now so this does not turn into cruft/rot as an unmaintained open (unsolicited, hahaha) PR. From other conversations I've read/silence here I am assuming it is not of interest, but I need it and I will continue to develop and maintain it on my fork and if you ever want it adapted back into upstream, happy to do the work, just LMK.

PR #102, opened 2026-09-01
Preserve per-identity thinker disabled state

24601:
## Problem

A bundled `disabled` marker is a safe default for a thinker that ships disabled, but treating it as shared code caused updates to overwrite per-identity enablement. Symlink installs could silently re-disable an explicitly enabled thinker, while fresh dashboard installs could omit a bundled disabled default.

## Fix

Treat `disabled` as per-identity state:

- fresh CLI and dashboard installs inherit the disabled default only when that bundled thinker ships a `disabled` marker;
- legacy symlinked markers are converted once to regular identity-local files;
- later syncs preserve either state, including an absent marker after explicit opt-in;
- docs describe dashboard/manual enablement and update durability.

This does not globally enable retrieval, restart identities, or change runtime settings.

## Verification

- `bash tests/test_identity_sync_thinkers.sh` — 26 passed, 0 failed


nickjalbert:
LGTM!  Thanks for the PR!  🚢

PR #103, opened 2026-09-01
Keep thinker shellm config private and reliable

24601:
## Summary

Keep thinker-launched shellm configuration both private and reliable across nested runs:

- normalize an inherited `LLM_API_URL` into shellm's effective endpoint configuration;
- let nested shellm inherit endpoint configuration instead of repeating it as `--var`, preserving shellm's Docker loopback rewrite;
- parent-export provider keys and skill-declared variables before process-substitution flag assembly, then forward their names rather than values;
- redact credential, endpoint, URL/URI, and DSN values from durable shellm command records and bug-report bundles.

## Why

Provider keys and skill configuration may be shell-local rather than already exported. Exporting them in the monolith parent before building nested shellm flags makes those runs work, while bare-name forwarding keeps values off process arguments. Endpoint inheritance also avoids a duplicate variable overriding shellm's safe Docker rewrite.

The bug-report scrubber sends each literal replacement program to `sed` through an anonymous process-substitution file descriptor, not argv or a named script. Per-literal isolation preserves later redactions after malformed values, while staging-local temporary outputs and subshell-local cleanup traps cover errors and signals without clobbering the enclosing bug-report cleanup. Underscore-delimited private-name matching covers endpoint and DSN values without misclassifying `CURL_OPTS`.

## Verification

- `bash tests/test_thinker_env_fallback.sh` — 10 passed, 0 failed
- `bash tests/test_var_secrets.sh` — 10 passed, 0 failed
- `bash tests/test_persona_bugreport.sh` — 54 passed, 0 failed
- `bash tests/test_llm_openai_compatible.sh` — 21 passed, 0 failed

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ life, and opens a dashboard where you can watch its mind run:
 curl -fsSL https://headlong.ai/install.sh | bash
 ```
 
-You'll need bash 3.2+, git, curl, jq, and an LLM API key (Anthropic,
+You'll need bash 3.2+, git, curl, jq, Python 3, and an LLM API key (Anthropic,
 OpenAI, Gemini, or OpenRouter) — or a local model on any
 OpenAI-compatible server (llama.cpp, Ollama, vLLM, LM Studio; see
 [Local models](#local-models) below, no key needed); the dashboard also
--- a/bin/shellm
+++ b/bin/shellm
@@ -2691,6 +2691,12 @@ exit \$__shellm_rc
         for _ev in "${_SHELLM_EXTRA_VARS[@]+"${_SHELLM_EXTRA_VARS[@]}"}"; do
             env_vars+=("$_ev")
         done
+        # Endpoint aliases are canonical shellm state, not ordinary extras.
+        # Keep the Docker-rewritten value last so a duplicate --var cannot
+        # restore a host-only URL inside generated code.
+        if [[ -n "$execution_api_url" ]]; then
+            env_vars+=(SHELLM_API_URL="$execution_api_url" LLM_API_URL="$execution_api_url")
+        fi
         if [[ "$SHELLM_DOCKER_ACCESS" == "broker" && -n "${_SHELLM_DOCKER_BROKER_DIR:-}" ]]; then
             env_vars+=(SHELLM_DOCKER_BROKER="$_SHELLM_DOCKER_BROKER_DIR")
             env_vars+=(SHELLM_DOCKER_TRANSPORT="$_SHELLM_DOCKER_BROKER_TRANSPORT")
--- a/bin/shellm
+++ b/bin/shellm
@@ -11,6 +11,20 @@ set -euo pipefail
 # a parent shellm run) always wins over files — same semantics as
 # thinkers/_lib/common.sh:_load_env_defaults. Values are extracted by
 # sourcing the file in a subshell so quoting behaves like a plain `source`.
+# Preserve endpoint values from the original process separately: the two
+# endpoint names are aliases, so a process LLM_API_URL must also beat a
+# SHELLM_API_URL subsequently loaded from a file (and vice versa).
+_shellm_process_shellm_api_url_set=0
+_shellm_process_llm_api_url_set=0
+if [[ -n "${SHELLM_API_URL+x}" ]]; then
+    _shellm_process_shellm_api_url_set=1
+    _shellm_process_shellm_api_url="$SHELLM_API_URL"
+fi
+if [[ -n "${LLM_API_URL+x}" ]]; then
+    _shellm_process_llm_api_url_set=1
+    _shellm_process_llm_api_url="$LLM_API_URL"
+fi
+
 _shellm_load_env() {
     local envfile="$1"
     [[ -f "$envfile" ]] || return 1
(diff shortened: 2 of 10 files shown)


nickjalbert (changes requested):
Thanks for the PR!  Two changes requested from review:

> 1. **Honor process endpoint values over values loaded from files** ([bin/shellm:73](https://github.com/24601/headlong/blob/c47d0b48c1695e81da52c28fde2422f07069fcc3/bin/shellm#L73))
  >
  > The environment loaders run before `SHELLM_API_URL` and `LLM_API_URL` are resolved as aliases. When the process supplies `LLM_API_URL` and `.env` supplies
  `SHELLM_API_URL`, line 73 selects the file value even though Headlong documents that process values take priority over file values.
  >
  > I reproduced the problem with `LLM_API_URL=http://operator.example.invalid/v1` in the process and `SHELLM_API_URL=http://file.example.invalid/v1` in the
  working directory's `.env`. Generated code received the file URL under both variable names. Shellm can therefore send prompts and credentials to a stale
  endpoint even though the operator supplied a different endpoint for the process.
  >
  > Please preserve whether each value came from the original process, then resolve process values before values loaded from files. Add a regression test with
  conflicting process and file URLs, and verify that both generated variables contain the process URL.

  > 2. **Keep broad matching for credential names without underscores** ([bin/shellm:2911](https://github.com/24601/headlong/blob/c47d0b48c1695e81da52c28fde2422f07069fcc3/bin/shellm#L2911-L2915), [tools/persona:663](https://github.com/24601/headlong/blob/c47d0b48c1695e81da52c28fde2422f07069fcc3/tools/persona#L663-L667))
  >
  > The new boundary rule no longer recognizes credential names such as `SERVICE_APIKEY`, `APIKEY`, `ACCESSTOKEN`, and `PGPASSWORD`. The previous substring
  check recognized these names.
  >
  > I reproduced the trajectory leak with a synthetic `PGPASSWORD` value. A `shellm --var PGPASSWORD=pg-secret-... task` invocation wrote the full value into
  the durable `shellm-run` row. The bug-report scrubber uses the same boundary rule, so it can retain compact credential values in an exported bundle.
  >
  > Please keep broad matching for credential terms such as `KEY`, `TOKEN`, `SECRET`, and password forms. Apply the stricter boundary rule only to endpoint
  terms such as `URL`, where it is needed to avoid false positives such as `CURL_OPTS`. Add compact credential cases to `tests/test_var_secrets.sh` and `tests/
  test_persona_bugreport.sh`, while keeping the existing `CURL_OPTS` regression case.

24601:
> Thanks for the PR! Two changes requested from review:
> 
> > 1. **Honor process endpoint values over values loaded from files** ([bin/shellm:73](https://github.com/24601/headlong/blob/c47d0b48c1695e81da52c28fde2422f07069fcc3/bin/shellm#L73))
> > 
> > The environment loaders run before `SHELLM_API_URL` and `LLM_API_URL` are resolved as aliases. When the process supplies `LLM_API_URL` and `.env` supplies
> > `SHELLM_API_URL`, line 73 selects the file value even though Headlong documents that process values take priority over file values.
> > I reproduced the problem with `LLM_API_URL=http://operator.example.invalid/v1` in the process and `SHELLM_API_URL=http://file.example.invalid/v1` in the
> > working directory's `.env`. Generated code received the file URL under both variable names. Shellm can therefore send prompts and credentials to a stale
> > endpoint even though the operator supplied a different endpoint for the process.
> > Please preserve whether each value came from the original process, then resolve process values before values loaded from files. Add a regression test with
> > conflicting process and file URLs, and verify that both generated variables contain the process URL.
> 
> > 2. **Keep broad matching for credential names without underscores** ([bin/shellm:2911](https://github.com/24601/headlong/blob/c47d0b48c1695e81da52c28fde2422f07069fcc3/bin/shellm#L2911-L2915), [tools/persona:663](https://github.com/24601/headlong/blob/c47d0b48c1695e81da52c28fde2422f07069fcc3/tools/persona#L663-L667))
> > 
> > The new boundary rule no longer recognizes credential names such as `SERVICE_APIKEY`, `APIKEY`, `ACCESSTOKEN`, and `PGPASSWORD`. The previous substring
> > check recognized these names.
> > I reproduced the trajectory leak with a synthetic `PGPASSWORD` value. A `shellm --var PGPASSWORD=pg-secret-... task` invocation wrote the full value into
> > the durable `shellm-run` row. The bug-report scrubber uses the same boundary rule, so it can retain compact credential values in an exported bundle.
> > Please keep broad matching for credential terms such as `KEY`, `TOKEN`, `SECRET`, and password forms. Apply the stricter boundary rule only to endpoint
> > terms such as `URL`, where it is needed to avoid false positives such as `CURL_OPTS`. Add compact credential cases to `tests/test_var_secrets.sh` and `tests/   test_persona_bugreport.sh`, while keeping the existing `CURL_OPTS` regression case.

Thx for the feedback, will add shortly

24601:
> Thanks for the PR! Two changes requested from review:
> 
> > 1. **Honor process endpoint values over values loaded from files** ([bin/shellm:73](https://github.com/24601/headlong/blob/c47d0b48c1695e81da52c28fde2422f07069fcc3/bin/shellm#L73))
> > 
> > The environment loaders run before `SHELLM_API_URL` and `LLM_API_URL` are resolved as aliases. When the process supplies `LLM_API_URL` and `.env` supplies
> > `SHELLM_API_URL`, line 73 selects the file value even though Headlong documents that process values take priority over file values.
> > I reproduced the problem with `LLM_API_URL=http://operator.example.invalid/v1` in the process and `SHELLM_API_URL=http://file.example.invalid/v1` in the
> > working directory's `.env`. Generated code received the file URL under both variable names. Shellm can therefore send prompts and credentials to a stale
> > endpoint even though the operator supplied a different endpoint for the process.
> > Please preserve whether each value came from the original process, then resolve process values before values loaded from files. Add a regression test with
> > conflicting process and file URLs, and verify that both generated variables contain the process URL.
> 
> > 2. **Keep broad matching for credential names without underscores** ([bin/shellm:2911](https://github.com/24601/headlong/blob/c47d0b48c1695e81da52c28fde2422f07069fcc3/bin/shellm#L2911-L2915), [tools/persona:663](https://github.com/24601/headlong/blob/c47d0b48c1695e81da52c28fde2422f07069fcc3/tools/persona#L663-L667))
> > 
> > The new boundary rule no longer recognizes credential names such as `SERVICE_APIKEY`, `APIKEY`, `ACCESSTOKEN`, and `PGPASSWORD`. The previous substring
> > check recognized these names.
> > I reproduced the trajectory leak with a synthetic `PGPASSWORD` value. A `shellm --var PGPASSWORD=pg-secret-... task` invocation wrote the full value into
> > the durable `shellm-run` row. The bug-report scrubber uses the same boundary rule, so it can retain compact credential values in an exported bundle.
> > Please keep broad matching for credential terms such as `KEY`, `TOKEN`, `SECRET`, and password forms. Apply the stricter boundary rule only to endpoint
> > terms such as `URL`, where it is needed to avoid false positives such as `CURL_OPTS`. Add compact credential cases to `tests/test_var_secrets.sh` and `tests/   test_persona_bugreport.sh`, while keeping the existing `CURL_OPTS` regression case.

Thx again for the detailed review, I've now addressed and made both requested changes.

1. Process-provided endpoint values are now captured before env-file loading and take precedence when the aliases are resolved. The regression test covers a process `LLM_API_URL` conflicting with a file-provided `SHELLM_API_URL` and verifies that generated code receives the process URL under both aliases.
2. Credential terms (`KEY`, `TOKEN`, `SECRET`, and password variants) retain broad matching, including compact names such as `SERVICE_APIKEY`, `APIKEY`, `ACCESSTOKEN`, and `PGPASSWORD`. Boundary matching is limited to endpoint terms, preserving the `CURL_OPTS` case.

While I was at it, I also hardened the related paths after an additional review: multiline and space-containing secrets are scrubbed byte-safely, bug-report generation now fails closed if redaction fails, and skill-declared endpoint aliases can no longer override shellm’s canonical Docker-rewritten endpoint.

LMK if there's anything else I can do here.

nickjalbert:
LGTM!  Thanks for the PR!

PR #106, opened 2026-09-02
persona: simple chat loop exits cleanly

dbreunig:
## Summary

Leaving the simple `persona` chat loop failed with `watcher: unbound variable` and left the background message watcher running. This fixes the EXIT trap in `tools/persona` and adds a regression test.

## Details

The simple chat loop runs when `headlong-tui` is not on PATH. You reach it by typing the identity's name alone, e.g. `iota`. The loop starts a background process that polls for new replies and prints them. It then registers an EXIT trap to kill that process when you leave.

The trap read a variable declared local to the chat function. The trap runs after the function has returned, so by then the variable is gone. Under `set -u` that is a fatal error. Leaving the chat by any route hit it: Ctrl-C, Ctrl-D, or a failed `chat send`. The user saw an error such as:

```
/root/.local/bin/iota: line 1: watcher: unbound variable
```

The watcher was never killed, and the error hid whatever caused the exit.

The fix is one line. The trap now uses double quotes, so bash writes the pid into the trap text at the moment it is set. The trap no longer depends on the variable.

`tests/test_persona_chat_exit.sh` builds a throwaway install, runs the simple chat loop with one piped line and then EOF, and checks four things. The chat exits 0. Nothing on stderr mentions an unbound variable. The watcher is gone afterward. The typed message landed in the trajectory. Before the fix, three of the four fail. The test passes under bash 5.3 and the macOS bash 3.2.

### DIFF
--- a/tests/test_persona_chat_exit.sh
+++ b/tests/test_persona_chat_exit.sh
@@ -1,0 +1,76 @@
+#!/usr/bin/env bash
+# tests/test_persona_chat_exit.sh — the simple `persona <name>` chat loop
+# (used when the Rust TUI is not installed) exits cleanly and takes its
+# background message watcher with it.
+#
+# Usage: tests/test_persona_chat_exit.sh
+#
+# Why: the loop's EXIT trap kills a watcher whose pid lives in a function
+# local. The trap runs after the function has returned, so it must carry the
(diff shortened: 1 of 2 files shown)


nickjalbert:
LGTM!  Added a small fix for shellcheck CI.

PR #108, opened 2026-09-04
slack: prepend capped prior lines on first @mention in a thread

headlong42:
## Why

A first `@mention` in an existing Slack thread currently delivers only that one event. The unaddressed history sitting above it never reaches the mind — that's the gap that produced the overclaim about already having the thread.

Nick asked whether the bridge should forward the whole thread once the bot is added to it. Yes, capped so a long thread cannot dump.

## What

On first join of an existing thread (`app_mention` / in-text mention, thread already has a parent `thread_ts`, thread not yet in `ActiveThreads`), the inbound **drain worker** (not the Bolt handler) calls `conversations.replies` and prepends a digest of earlier lines:

```
(thread before this mention, last N)
  display-name: text
  ...
```

then the mention itself, same `from_name` / permalink as today.

- Cap: `SLACK_THREAD_JOIN_BACKFILL` (default 20, max 50, `0` disables).
- Fail open: Slack timeout / error still delivers the mention with no digest.
- Already-active threads, top-level mentions that *start* a thread, and DMs skip the lookup.
- Bolt still acks immediately.

No new Slack scopes (`conversations.replies` is covered by the existing history / channels read scopes).

## Not in this PR

Outbound `reactions.add` (sending emoticons). Receiving reactions is #88. Sending them is a new `chat` surface plus `reactions:write` + reinstall — not a small add. Happy to do it as a follow-up if you want it.

## Tests

`uv run --project slack pytest slack/tests/` — 55 passed.

### DIFF
--- a/slack/README.md
+++ b/slack/README.md
@@ -11,6 +11,8 @@ are posted back to the right conversation.
 Slack <=(Socket Mode websocket)=> headlong-slack-bridge
     inbound:  event -> from_name "slack-<user>-<channel>[-<thread_ts>]"
               -> POST <web>/api/identities/<id>/chat
+              first @mention in an existing thread prepends a capped
+              prior-line digest (conversations.replies, fail-open)
               reaction_added on the bot's messages (and any DM reaction)
               lands as a `:emoji:` body (same from_name / thread)
     outbound: tail trajectory.jsonl -> message steps where from=<identity>
@@ -53,7 +55,9 @@ runs its own Slack-connected agent on.
 Other settings: `HEADLONG_SLACK_STATE_DIR`, legacy `SHELLM_SLACK_STATE_DIR`
 (cursor + thread state, default
 `<identity>/run/slack-bridge/`), `SLACK_THREAD_FOLLOWUPS=1` (answer
-un-mentioned replies in threads the bot is already part of).
+un-mentioned replies in threads the bot is already part of),
+`SLACK_THREAD_JOIN_BACKFILL` (how many messages above a first @mention
+to prepend; default 20, 0 disables, max 50).
 
 For the end-to-end procedure we used to install our agent, Audel, into a
 workspace on that stack (Slack app, tokens, SSM env, rebuild, verification),
--- a/slack/README.md
+++ b/slack/README.md
@@ -56,7 +56,7 @@ Other settings: `HEADLONG_SLACK_STATE_DIR`, legacy `SHELLM_SLACK_STATE_DIR`
 (cursor + thread state, default
 `<identity>/run/slack-bridge/`), `SLACK_THREAD_FOLLOWUPS=1` (answer
 un-mentioned replies in threads the bot is already part of),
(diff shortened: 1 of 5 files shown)


nickjalbert (changes requested):
Thanks!  Two requested fixes from review:


> 1. **The cap selects the oldest replies instead of the latest context** ([slack/src/headlong_slack/inbound.py:334](https://github.com/laude-institute/headlong/blob/48a706cd3bca8fea65ee664b4295aec4c1eea79a/slack/src/headlong_slack/inbound.py#L334-L341))
>
> Slack returns `conversations.replies` time pages in earliest first order. When a thread has more messages than `fetch`, this request returns the parent and the earliest replies before the mention. Taking the tail of that page cannot recover the replies closest to the mention. In a reproduction with a two message cap, a parent, and four earlier replies, the function returned the first two replies instead of the last two. The current test misses the failure because its fake returns every configured message without applying the requested `limit`.
>
> Please make the lookup return the closest prior messages while preserving the text cap and a bounded failure path. Use cursor pagination or another documented Slack API approach, and add a test whose fake applies `limit` to a thread longer than the configured cap. The test should assert that the two replies immediately before the mention are included.
>
> 2. **Timed out backfills leave stale API calls queued** ([slack/src/headlong_slack/inbound.py:343](https://github.com/laude-institute/headlong/blob/48a706cd3bca8fea65ee664b4295aec4c1eea79a/slack/src/headlong_slack/inbound.py#L343-L352))
>
> `Future.result(timeout=...)` stops waiting but does not stop the submitted call. Each later first mention submits another lookup to the same one worker executor. In a reproduction with five slow lookups, all five deliveries timed out and failed open, but every queued `conversations.replies` call still ran afterward. A slow Slack call can therefore leave a growing stale queue, occupy the lookup worker, and spend API quota after the related messages have already been delivered.
>
> Please keep at most one backfill lookup in flight, as the existing reaction lookup path does, or reject new lookup work while the shared worker is busy. Add a delayed call test that sends a burst of first mentions, confirms only one Slack lookup runs, and confirms a normal inbound message is delivered without waiting for the stale calls.

nickjalbert (changes requested):
One more fix request:

> 1. **The timeout regression test still passes with the queueing bug** ([slack/tests/test_inbound.py:736](https://github.com/laude-institute/headlong/blob/d5cc776ba1ba2837c73a739832752347c1f2d69b/slack/tests/test_inbound.py#L730-L747))
>
> The test checks `client.replies_calls` before the first 0.8 second fake Slack call finishes. The fake increments that list only when a queued call starts, so the count is still one even if four more calls are already queued. Cleanup then calls `shutdown(cancel_futures=True)` and cancels those queued calls before they can change the count. I replaced `_try_lookup` at runtime with the old behavior that submits every lookup, and this exact regression test still passed in 0.31 seconds.
>
> Please make the test observe submissions directly, or wait until the first slow call has finished before asserting that no second call starts. Verify that the test fails when `_try_lookup` always submits work and passes with the current one in flight guard. The production fix itself worked in the focused reproduction.

PR #109, opened 2026-09-05
slack: upload chat send-file steps via files.upload_v2

headlong42:
Outbound used to `chat.postMessage` every reply, so a `chat send-file` step to a `slack-*` name never actually uploaded.

This slice:
- detects file-bearing mind-log steps (`filename` + `content` / `content_b64`, same contract as Telegram)
- uploads via `files.upload_v2` into the right channel/thread
- falls back to a short notice if decode/upload fails, without killing later replies
- adds bot scope `files:write` in `manifest.json`

Needs a Slack app reinstall / re-grant for `files:write` before uploads will land (same drill as `reactions:read`).

### DIFF
--- a/slack/PLAYBOOK.md
+++ b/slack/PLAYBOOK.md
@@ -40,9 +40,10 @@ After the rebuild, copy the archive back and run `identity import`.
 
 Work from the `slack/` directory. `manifest.json` is the source of truth.
 
-`reaction_added` needs bot scope `reactions:read` (in `manifest.json`). A
-scope change requires reinstall / re-grant; until then Socket Mode will
-not deliver reaction events.
+`reaction_added` needs bot scope `reactions:read` (in `manifest.json`).
+Outbound `chat send-file` needs `files:write`. A scope change requires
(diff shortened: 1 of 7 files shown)


nickjalbert (changes requested):
From review:

> 1. **Do not suppress a retry after Slack rejects the first upload** ([slack/src/headlong_slack/outbound.py:117](https://github.com/headlong42/headlong/blob/692e16e26ee2def40cbf4fc79ebbcc61e60ffd16/slack/src/headlong_slack/outbound.py#L117-L130))
>
> `recent.is_duplicate(...)` records the file signature before `files_upload_v2` runs. If Slack has a transient failure and the agent repeats the same `chat send-file` step, the first step posts a failure notice, but the second step is skipped for five minutes without another upload attempt. I reproduced the case with two identical file steps and a client that fails once. The bridge made one upload attempt and posted one failure notice.
>
> Record the signature only after Slack accepts the upload, or remove the recorded signature when the upload fails. Add a regression test where the first identical upload raises and the second succeeds, then rerun `uv run --project slack pytest slack/tests`.

PR #111, opened 2026-09-05
slack: add outbound emoji reactions via chat react

headlong42:
Nick asked for outbound Slack emoji reactions (respond with emojis).

`chat react <slack-name> <emoji>` stamps a chat-sourced message step with a Slack short name (`thumbsup`, `eyes`, `+1`). The Slack bridge calls `reactions.add` on the inbound permalink (`source_url`) when present, otherwise the conversation `thread_ts`. Telegram skips those steps so they are not posted as text.

Needs bot scope `reactions:write` — reinstall after merge/deploy.

Usage:
```bash
chat react slack-U07AB12CD-C09XYZ123-1722400000.123456 thumbsup
chat react --reply-to <inbound-step-id> slack-U07AB12CD-C09XYZ123-1722400000.123456 eyes
```

Tests: slack 85, telegram 71, `tests/test_chat_react.sh` 8/8.

### DIFF
--- a/bin/chat
+++ b/bin/chat
@@ -263,6 +263,63 @@ cmd_send_file() {
     printf 'File sent: %s\n' "$filename" >&2
 }
 
+cmd_react() {
+    local from="${IDENTITY_NAME:-}" to="" reply_to="" source_url="" emoji=""
+    while [[ $# -gt 0 ]]; do
+        case "$1" in
+            --from) from="${2:?--from requires a name}"; shift 2 ;;
+            --to)   to="${2:?--to requires a name}"; shift 2 ;;
+            --reply-to) reply_to="${2:?--reply-to requires a step id}"; shift 2 ;;
+            --source-url) source_url="${2:?--source-url requires a URL}"; shift 2 ;;
+            *) break ;;
+        esac
+    done
+    if [[ -z "$to" && $# -gt 0 ]]; then
+        to="$1"; shift
+    fi
+    [[ -z "$to" ]] && die "Usage: chat react [--reply-to <step_id>] [--source-url <url>] <to_name> <emoji>"
+    while [[ $# -gt 0 ]]; do
+        case "$1" in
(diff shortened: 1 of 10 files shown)


nickjalbert (changes requested):
One requested fix from review:

> 1. **Make emoji validation work on every supported Bash 3.2 build** ([bin/chat:298](https://github.com/laude-institute/headlong/blob/6a98ab5946bd82d6d23436f0a7a3600dea18aaa8/bin/chat#L298))
>
> The escaped hyphen in `^[a-z0-9_+\-:]{1,64}$` is not portable across the Bash 3.2 environments that Headlong supports. In the macOS CI job, `chat react` rejects `thumbsup`, `eyes`, `+1`, and `white_check_mark`, so five positive cases fail and the new command does not work there.
>
> Put the hyphen at the start or end of the bracket expression, where it does not need an escape, for example `^[-a-z0-9_+:]{1,64}$`. Then rerun `tests/test_chat_react.sh` with Bash 3.2 first on `PATH`, and rerun the complete CI suite. The macOS Bash 3.2 job and the eight focused reaction cases must pass.

headlong42:
Addressed the Bash 3.2 emoji-class review: hyphen is now first in the bracket expression (`^[-a-z0-9_+:]{1,64}$`) so it does not need an escape.

Pushed on `audel/slack-outbound-reactions` as 1d5670a.

Verified here:
- `tests/test_chat_react.sh` — 8/8 pass (Bash 5.2)
- slack pytest — 85 passed
- telegram pytest — 71 passed

I do not have Bash 3.2 on this host (no compiler to build 3.2.57), so the macOS CI job is the remaining 3.2 check. Full `tests/run-all.sh` was still running and had unrelated pre-existing FAILs (llm ledger path, persona chat, install PATH, etc.) that are outside this one-line change.

nickjalbert:
LGTM, thanks!

PR #112, opened 2026-09-05
traj: formatted tail and cat return the steps --filter matches

MaxFreedomPollard:
Follow-up to #110, same function.

## What

In formatted mode — what a terminal gets by default — `traj tail` and
`traj cat` end the stream at the first step `_format_line` cannot turn into
six lines, exit 1, and say nothing. Two ways to hit that:

**A filter.** `--filter type=thought` at a terminal prints nothing at all,
because the first step of every trajectory is its header and no type filter
matches it. Widen the filter so the header does match and it becomes silent
truncation instead:

```
$ traj cat --filter type=trajectory,thought        # at a tty
2026/09/05 03:25:48 e0e9ef46 [trajectory] root
2026/09/05 03:25:48 a882d8db [thought] alpha
$ echo $?
1                                                    # gamma, also a thought, is gone
```

**A step the preview cannot render.** `traj append` takes JSON on stdin
(`thinkers/monolith/prompt.md`) and checks only that it parses, so `.content`
can be an object. The preview slices it, jq errors after one of its six
lines, and the stream dies there — `-r` included, because its pre-filter
passes the step. A malformed line does the same to a bare `traj cat` with no
flags. `-a` gets past all of these — it renders the object step and skips the
malformed line — and `--raw` returns every step; the default formatted view is
the one that dies, and `-r` dies with it on the unrenderable step.

`tail -f --filter …` is the sharpest form: the follower dies at the first
non-matching step and stays dead, with no diagnostic.

## Why

`_format_line` hands each step to a jq program that begins with
`select(matches_filters($filters))` and reads the six result lines back with a
`{ read; read; … } < <(jq …)` group. Whenever jq emits fewer than six lines —
a filtered-out step emits none, an unrenderable one emits one — a `read` fails
at EOF, and under the file's `set -euo pipefail` that kills the `while read`
subshell driving the stream. The guard meant to skip such a step,
`[[ -n "$entry_type" ]] || return 0`, is right after the group and is never
reached.

2c74855 states the invariant this breaks — *"a single corrupt line skips
gracefully instead of aborting and dropping all subsequent entries"* — and
hardened every jq call that reads a trajectory file except this one, which is
handed a single line and cannot use `fromjson?`. 93ad7bb then added `--filter`,
which made the dead guard matter for every non-matching step rather than only
for corrupt lines.

The `-a` branch of the same function already does this right: capture jq's
output, `|| return 0`, guard empty, then split. This makes the default branch
do the same.

## The fix

```diff
-    local entry_type content source ts step_id traj_hex
-    { read -r entry_type
-      …
-    } < <(
-        printf '%s' "$line" | jq -r … '
+    local _fields
+    _fields=$(printf '%s' "$line" | jq -r … '
             …
-        ' 2>/dev/null
-    )
+        ' 2>/dev/null) || return 0
+    [[ -n "$_fields" ]] || return 0
+
+    local entry_type content source ts step_id traj_hex
+    { read -r entry_type
+      …
+    } <<< "$_fields" || true
     [[ -n "$entry_type" ]] || return 0
```

The jq program is unchanged. Two details worth a sentence each:

- **Why capture rather than `|| true` on the old group.** A one-token
  `|| true` after the process substitution also stops the deaths and passes
  every filter case — but on an unrenderable step it reads the one line jq
  did emit and prints a fabricated `----/--/-- --:--:--  [thought]` line with
  a blank id. Capturing first lets `|| return 0` see jq's exit status and skip
  the step. The test pins the difference.
- **Why `|| true` on the new group.** `$(…)` drops trailing newlines, so a
  record whose last field is the empty string comes back a line short. The
  process substitution kept that newline and read the field as empty; this
  keeps doing so instead of failing on the sixth read.

`_format_line` is reached only in formatted mode. Every scripted `traj` call
in the repo passes `--raw` and is untouched. One automated caller does run
formatted: the TUI sets `CLICOLOR_FORCE=1` on every `traj` it spawns
(`tui/headlong/src/main.rs:341`) and auto-refreshes with `tail -r -n 50`. For
any tree that view renders today its output is byte-identical; where it
used to die on an unrenderable step and show a truncated tree, it now shows
the rest. A `tail --filter …` typed into its `/traj` pane hits this bug today.
Output for every step that rendered before is unchanged; two small
differences, both matching the `-a` branch: a step whose preview jq errors is
now skipped rather than ending the stream, and on bash ≥ 4.4 a NUL byte in a
step's content now draws bash's "ignored null byte" warning on stderr, as it
already does under `-a`.

## Tests

`tests/test_traj_formatted_filter.sh` pins, in formatted mode: `cat` and
`tail` with a type filter exit 0 and print the matching steps; formatted and
raw modes select the same step ids (raw is the oracle); a filter only the
header matches prints one line and exits 0 — the first non-match no longer
ends the stream; a filter nothing matches prints nothing and exits 0; two
`--filter` flags AND and comma values OR; the `-a` and raw neighbours are
unchanged; a formatted line still carries stamp, id, `[type, source]` and
content in that order, so the six reads stay matched to the six lines jq
emits; `-r` still prefixes the id with the file hex; a malformed line no
longer truncates the stream and leaks nothing to stderr; a record with an
empty trailing field is shown; and a step jq cannot render is skipped with no
fabricated line in its place.

On `main`: 9 passed, 17 failed — the nine are the neighbour and rendering
controls. With the fix: 26 passed, 0 failed. The one-token alternative above:
25 passed, 1 failed, on the fabricated-line check. Mutation-tested against
fourteen deliberate breakages of `bin/traj` — reverting to the process
substitution, dropping the capture's `|| return 0`, inverting the empty guard,
dropping the read group's `|| true`, ignoring the filter, printing only the
first match, hardcoding the type, reordering or swapping jq fields, dropping a
read, breaking the timestamp branch, leaking jq's stderr, corrupting the
content column — thirteen are caught. The fourteenth, deleting
`[[ -n "$_fields" ]] || return 0`, is not observable because the read group
tolerates a short record and the `entry_type` guard then returns; the line
stays because it mirrors the `-a` branch and says what is meant.

Checked locally: `shellcheck -S warning` clean over the CI file set, the new
test clean at `-S info` too. Parses and runs under Apple's bash 3.2 and a pure
BSD userland as well as GNU coreutils; stable across timezones and locales.
`cloc bin/ thinkers/` +1 code line. `tests/run-all.sh` green apart from two
failures that are pre-existing on `main` here and green in CI:
`test_persona_chat_exit.sh` (pins `PATH` to `/usr/bin:/bin`, which has no `jq`
on a Homebrew Mac) and `test_workspace_runtime.sh` (`_runtime_line` stats
`.git/HEAD`, which does not exist when the checkout is a git worktree, as mine
is).

### DIFF
--- a/bin/traj
+++ b/bin/traj
@@ -958,15 +958,15 @@ _format_line() {
         return 0
     fi
 
-    local entry_type content source ts step_id traj_hex
-    { read -r entry_type
-      IFS= read -r content
-      read -r source
-      read -r ts
-      read -r step_id
-      read -r traj_hex
-    } < <(
-        printf '%s' "$line" | jq -r --argjson filters "$_filter_json" '
+    # Capture first, then guard, the way the -a branch above does. A step the
+    # filter drops makes jq print nothing, and reading six lines straight from
+    # an empty process substitution failed at the first read; under set -e
+    # that ended the whole formatted stream instead of skipping one step, so
+    # `traj tail --filter type=...` at a terminal printed nothing and exited
+    # 1, usually on the header line. The same read cut an unfiltered stream
+    # off at the first malformed line.
+    local _fields
+    _fields=$(printf '%s' "$line" | jq -r --argjson filters "$_filter_json" '
             def matches_filters($fs):
                 . as $step |
                 ($fs | length == 0) or all($fs[];
@@ -994,8 +994,21 @@ _format_line() {
             ((.ts // .timestamp // "0") | tostring),
             (.step_id // "-"),
             (._traj_hex // "-")
-        ' 2>/dev/null
-    )
+        ' 2>/dev/null) || return 0
+    [[ -n "$_fields" ]] || return 0
+
+    local entry_type content source ts step_id traj_hex
+    # The capture drops trailing newlines, so a record whose last field is the
+    # empty string arrives one line short. The process substitution this
+    # replaces kept that newline and read the field as empty; keep doing so
+    # rather than letting the sixth read fail and end the stream.
+    { read -r entry_type
+      IFS= read -r content
+      read -r source
+      read -r ts
+      read -r step_id
+      read -r traj_hex
+    } <<< "$_fields" || true
     [[ -n "$entry_type" ]] || return 0
 
     # Normalize placeholders back to empty
--- a/tests/test_traj_formatted_filter.sh
+++ b/tests/test_traj_formatted_filter.sh
@@ -1,0 +1,155 @@
+#!/usr/bin/env bash
+# tests/test_traj_formatted_filter.sh — `traj tail --filter` and `traj cat
+# --filter` work in formatted mode, which is what a terminal gets.
+#
+# Usage: tests/test_traj_formatted_filter.sh
+#
+# Why: in formatted mode _format_line hands each step to jq, whose program
+# starts with `select(matches_filters($filters))`, and reads the six result
+# lines back with a `{ read; read; ... } < <(jq ...)` group. When a step does
+# not match, jq emits nothing, the first `read` fails at EOF, and under the
+# file's `set -e` that kills the `while read` subshell driving the stream. The
+# guard meant to skip a non-matching step, `[[ -n "$entry_type" ]] || return 0`,
+# sits right after the group and is never reached. So `traj tail --filter ...`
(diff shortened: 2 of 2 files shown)


nickjalbert:
LGTM; thanks!!

Issue #117, opened 2026-09-10
tests/run-all.sh: identity new writes test identities into the caller's live .identities when run from an activated identity


custos-1f916:
## Symptom

`tests/run-all.sh` run while `IDENTITY_NAME`/`IDENTITY_DIR` are set (i.e. from inside an activated identity) can create test identities (ada, alpha, the `chatexit*` fixtures, a `default` symlink) inside the **live** identity's `.identities` directory.

## Real-world impact (2026-09-10)

Upstream `tests/run-all.sh` was executed from inside an activated headlong identity. It left six stray identities in the live `.identities` directory. The fixtures held no memories, keys or goals and no model call ever ran against them; they were audited and removed. Harm was zero, but the fixture lived in the same room as the thing under test — exactly the class of leak an isolated HOME is supposed to prevent.

## Repro

```bash
git clone --depth 1 https://github.com/laude-institute/headlong /tmp/hl && cd /tmp/hl
identity new testfoo --default          # or any activation; exports IDENTITY_NAME/IDENTITY_DIR
source <(echo "IDENTITY_NAME=whatever") # minimal: the three vars below
bash tests/run-all.sh
ls .identities/                         # see testfoo, ada, alpha, chatexit*, default
```

Minimal env needed: `IDENTITY_NAME` + `IDENTITY_DIR` pointing at a dir containing `info.txt` (plus `TRAJ_ID`/`TRAJ_DIR` as an activated wake exports them).

## Root cause

`tools/identity` (commit 5458fee, lines 16–25) resolves its identities **root** from the inherited identity env:

```bash
if [[ -n "${IDENTITY_NAME:-}" && -n "${IDENTITY_DIR:-}" && -f "${IDENTITY_DIR}/info.txt" ]]; then
    _ACTIVE_IDENTITY_DIR="$IDENTITY_DIR"
    IDENTITY_DIR="$(dirname "$IDENTITY_DIR")"
else
    IDENTITY_DIR="${IDENTITY_DIR:-.identities}"
fi
```

The tests assume `identity new` lands in the test's own temporary app dir, and they vary `HOME` / `HEADLONG_APP_DIR` accordingly:

- `tests/test_status.sh:49` — `identity new ada --default` (inherits the caller's identity env)
- `tests/test_persona_bugreport.sh:70` — `identity new alpha`
- `tests/test_persona_chat_exit.sh:62` — `identity new "$NAME"` (multiple `chatexit*` names)

[...]

PR #118, opened 2026-09-10
tests: strip inherited identity/trajectory env from run-all.sh (#117)

custos-1f916:
## Symptom
Running `tests/run-all.sh` from inside an **activated identity** leaks `IDENTITY_NAME`/`IDENTITY_DIR` into the tests. `tools/identity` re-roots to the caller's live `.identities` in that state (see `tools/identity:20-22`), so `identity new` inside a test writes the `ada`/`alpha`/`chatexit*`/`default` fixtures into the **live** identity instead of the test's own temp app dir. This is the 2026-09-10 incident (#117): six stray identities left in a live `.identities`.

## Fix
`run-all.sh` unsets the identity + trajectory env at the top, so the suite is hermetic no matter what activated it. The unset list matches exactly the `deactivate` unset (`tools/identity:428`) plus the `TRAJ_ID`/`TRAJ_DIR`/`ROOT_TRAJ_ID` an activated wake exports. Every individual test unsets or exports its own values, so none of these are needed by the suite itself.

```bash
unset IDENTITY_NAME IDENTITY_DIR MEM_DIR SKILLS_DIR SKILLS_KERNEL_DIR TRAJ_ID TRAJ_DIR ROOT_TRAJ_ID
```

## Verification (ran from inside an activated identity)
- **Before (upstream)**: running the suite with `IDENTITY_NAME`/`IDENTITY_DIR` set wrote `ada`, `alpha`, `chatexit*`, `default` into the live `.identities`.
- **After (this PR)**: full suite run with the identity-activated env → **only `custos`** remains in the live `.identities`; `ada`/`alpha`/`chatexit*`/`default`/`testfoo` all land in each test's own temp `HOME`. No leak.
- Pre-existing red: `test_monolith_wake_sections.sh` has 2 failures (`matched memory listed`, `state file has related_prev`) that fail identically **standalone under a clean `env -i`** at this commit — unrelated to this change (my diff touches only `run-all.sh`). Flagged, not fixed here.

### DIFF
--- a/tests/run-all.sh
+++ b/tests/run-all.sh
@@ -12,6 +12,14 @@
 
 set -uo pipefail
 
+# Hermetic by design (laude-institute/headlong#117): running the suite from inside
+# an activated identity must not leak its identity/trajectory env into the tests.
+# tools/identity re-roots to the caller's live .identities when IDENTITY_NAME and
+# IDENTITY_DIR point at an active identity, so `identity new` inside a test would write
+# there instead of the test's own temp app dir. Every test unsets or exports its own
+# values, so the suite itself needs none of these.
+unset IDENTITY_NAME IDENTITY_DIR MEM_DIR SKILLS_DIR SKILLS_KERNEL_DIR TRAJ_ID TRAJ_DIR ROOT_TRAJ_ID
+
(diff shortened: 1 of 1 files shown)


nickjalbert:
Thanks!

PR #123, opened 2026-09-14
traj: add verify command with test

headlong42:
Adds a `traj verify` command that validates trajectory step signature chains using HMAC-SHA256.

**Changes:**
- Added `cmd_verify` in `bin/traj`: validates trajectory step signature chain using HMAC-SHA256
- Treats pre-signed lines as legacy (unverified) — returns 0 for verified/legacy
- Returns 0 for verified/legacy, 1 for tampered, 2 for errors
- Supports `--verbose` flag for per-step status
- Includes comprehensive test in `tests/test_traj_verify.sh` (27 tests pass)

**Note:** `traj verify` shells out to `jq` + `openssl` per line — ~1hr on 149K lines. Could optimize by batching crypto ops or document as batch job.

Commit: 1adc15c


custos-1f916:
CI was red on three jobs at head 1adc15ce (bash tests, macOS bash 3.2, gitleaks). Root causes and fixes are in the rebased follow-up PR (b086a46 on custos-1f916/headlong:pr123-ci-fix): prev_hash/sig need to be meta in bin/context's is_meta, the new test needs portable sed/openssl, and the test signing key needs a gitleaks allowlist. All three verified green locally (7/7, 27/27, no leaks).

PR #120, opened 2026-09-10
shellm: cap runaway command output with SHELLM_MAX_OUTPUT_SIZE

ousamabenyounes:
Add an output-size guard to bin/shellm's exec watchdog so a command producing unbounded, continuously-growing output (which keeps resetting the inactivity counter) is killed once its output exceeds SHELLM_MAX_OUTPUT_SIZE (default 100 MB, 0 disables). Fixes #116.

## Test verification (RED → GREEN)

With the fix reverted, the new test fails (RED):

```
FAIL runaway output is killed at SHELLM_MAX_OUTPUT_SIZE — [1;34m▶ Final answer received[0m

[90m  Trajectory: 82cbe4ab-0f4a-48ec-9a04-61be42e993a1[0m
FAIL kill feedback names the output-size cause

0 passed, 2 failed
```

With the fix applied, the test passes (GREEN):

```
GREEN attempt 1/2 (exit 0)
ok   runaway output is killed at SHELLM_MAX_OUTPUT_SIZE
ok   kill feedback names the output-size cause

2 passed, 0 failed
```

## Full local suite

Command: `bash tests/run-all.sh`

```
  ok   test_persona_chat_exit.sh (1s)
  ok   test_prompt_file.sh (2s)
  ok   test_prompt_section_failure.sh (0s)
  ok   test_recap_context_upgrade.sh (0s)
  ok   test_recap_render_cache.sh (1s)
  ok   test_recap.sh (2s)
  ok   test_recent_stream_filter.sh (1s)
  ok   test_redact_sensitive.sh (0s)
  ok   test_related_memories.sh (0s)
  ok   test_responder_deferral.sh (9s)
  ok   test_responder_idempotency.sh (0s)
  ok   test_responder_metrics.sh (4s)
  ok   test_responder_person_notes.sh (2s)
  ok   test_responder_thread_context.sh (1s)
  ok   test_retrieval_thinker.sh (2s)
  ok   test_rollup_model_precedence.sh (0s)
  ok   test_sandbox_gate.sh (1s)
  ok   test_sandbox_symlink_mounts.sh (1s)
  ok   test_shellm_context_scope.sh (5s)
  ok   test_shellm_empty_response_retry.sh (0s)
  ok   test_shellm_mount_coverage.sh (0s)
  ok   test_shellm_retry_feedback.sh (7s)
  ok   test_skills_split.sh (0s)
  ok   test_thinker_env_fallback.sh (2s)
  ok   test_thinkers_drain.sh (13s)
[...]

PR #127, opened 2026-09-17
shellm: ignore spurious SIGHUP in run_loop so the timeout feedback lands

### LINKED ISSUE #125: macOS (bash 3.2): test_inactivity_beacon.sh case 3 red on main since #120 — nested ceiling kill SIGHUPs the parent (rc=129) before the feedback step lands
## Bug

On the `macOS (bash 3.2)` CI job, `tests/test_inactivity_beacon.sh` case 3
("a nested run past the ceiling dies, and says why") fails on main. The nested
run *is* killed and the watchdog message lands (case 3's first assertion
passes), but the `type=feedback` trajectory step the model is supposed to see
("A nested shellm run was still alive") never gets written — the parent shellm
dies with `Hangup: 1` (SIGHUP, rc=129) before it can flush the feedback.

```
FAIL kill feedback names the sub-run, not an interactive prompt — shellm rc=129;
.../tests/test_inactivity_beacon.sh: line 73: 18999 Hangup: 1  $TIMEOUT "$WORK/toolbin/shellm" --workdir "$WORK/wd" --max-iterations 2 "$@"
```

## Evidence (main-side, not a branch)

| commit | run | result |
|--------|-----|--------|
| `c93048c9` | 35182692606 (04:38Z) | **green** |
| `5a63a63b` (merge of #120) | 35191994203 (06:55Z) | **red** |
| `03ffdf87` | 35194010121 (07:20Z) | **red** |

Two consecutive reds after the #120 merge, green before — consistent, not a flake.
The bash suite runs only under the macOS job, so this surfaces there.

## Suspected cause

#120 (`fix/issue-116`, SHELLM_MAX_OUTPUT_SIZE + stop-runaway-trees) added
`_kill_exec_tree` / `_supervise_exec` in `bin/shellm`, whose comment says they
"also reach nested shellm process groups." When the watchdog kills the nested
run's tree, the SIGHUP appears to reach the parent shellm's own process group on
macOS bash 3.2, so the parent is killed (rc=129) before it writes the feedback
step. Linux/bash-5 process-group semantics differ, so it's macOS-only.

## Repro

`bash tests/test_*.sh` (or run-all) under bash 3.2; case 3 of
`tests/test_inactivity_beacon.sh`. Bounded by `--max-iterations`; no model needed
(stubbed LLM).

## Environment

macOS CI runner, bash 3.2. Main `03ffdf87`.

custos-1f916:
## Summary

Fixes #125: `tests/test_inactivity_beacon.sh` case 3 red on macOS (bash 3.2) since #120.

## Root cause

On macOS bash 3.2, when the watchdog kills a nested exec tree (the STOP+KILL walk added in #120), a SIGHUP reaches the parent shellm's own process group. `run_loop` had no HUP trap, so the parent died (rc=129) before it could write the "A nested shellm run was still alive" feedback trajectory step. Linux/bash-5 process-group semantics differ, so only the macOS job goes red.

## Fix

`trap  HUP` at the top of `run_loop`. shellm is a non-interactive CLI (stdin=/dev/null); its intended stop signal is TERM (harness watchdog, `timeout`). A spurious SIGHUP must not drop the run before its timeout feedback is recorded. The supervisor's own HUP handling lives in the backgrounded `execute_code` subshell and is unaffected.

## Verification

Clean-env Linux (bash 5.2.15), all #120-touched suites:
- `tests/test_inactivity_beacon.sh`: 5 passed, 0 failed
- `tests/test_exec_termination.sh`: 2 passed, 0 failed
- `tests/test_output_size_watchdog.sh`: 5 passed, 0 failed

No Linux regression. The macOS job is the real confirmation (I cannot run bash 3.2 locally).


nickjalbert:
Should be fixed by #126.  Thanks though; lemme know if you still see the issue!

PR #128, opened 2026-09-17
Point the monolith error step at the run's last failing shell-output

custos-1f916:
## Failure-path certification: a failed no-durable-step run's receipt is now stranger-rerunnable

A failed no-durable-step run's receipt was a bare self-keyed `no durable step` claim; a stranger rerunning it could not see the actual diagnostic. The run's shell-output steps (the real diagnostic) are already in the trajectory, but the error step never referenced them.

This adds a `diagnostic_step` field to the run-failed error step, pointing at the run's **last failing shell-output** (the command that actually exited non-zero). When no command actually failed — e.g. an inference-layer death that leaves no failed shell — it falls back to the last non-empty shell-output. The fallback is provably never worse than the last-non-empty pick: it degrades to it exactly on the runs where no command failed. Absent when the run never ran a shell (nothing to reference — correct).

### Why the last *failing* step, not the last non-empty one
On 109 real orphaned failed runs, the run's last step is usually a benign status read (help text, a listing) that ran *after* the command that actually failed. Pointing at the last non-empty shell-output recovers the failing command in only 15.6% of those runs; pointing at the last failing shell-output recovers it in 63.3%. The real error was earlier in the run, and the last-failing pick finds it.

### Complementary to the existing `stderr_tail` field
`stderr_tail` is inline process stderr (last 3 lines, 300-char truncated); `diagnostic_step` is a **rerunnable pointer** to the durable shell-output step. In the discriminating case — a run whose diagnostic was a shell-output step rather than process stderr — only `diagnostic_step` recovers it. The two fields cover orthogonal axes: *what the process said* vs *what the run did*.

### Latent jq bug fixed
The old selector used `jq -Rr`, where each line is a separate input, so the program emitted one result per line and `tail -n 1` masked the rest. Now `jq -Rnr` + `inputs` collects all lines into one array before picking.

### Evidence
`tests/test_monolith_backoff.sh` **13/13 pass**, incl.:
- **6b** — failed run with a failing shell-output *and* a benign exit-0 step after it: error step carries `diagnostic_step=diag-so-1` (the failing step, not the later benign one) — discriminates the new selector from the old
- **6c** — bare failed run (no shell-output): error step has no `diagnostic_step` (nothing to reference)
- **6d** — failed run whose only shell-output has no exit code: `diagnostic_step` falls back to the last non-empty step (inference-layer death)

### Scope
2 commits, 2 files. No behavior change to the success path; only the run-failed error step gains an optional field.


nickjalbert:
LGTM thanks!

Issue #119, opened 2026-09-10
mem search can hang forever: the heartbeat defeats the inactivity watchdog and leaks orphaned processes


SimonvanAs:
### Symptom

`mem search` can hang indefinitely. The process stays alive printing its heartbeat, the run that started it never finishes, and the identity's mind stops advancing until something kills it by hand.

Observed on a long-running identity over two days:

- A monolith wake frozen ~20 minutes at backoff level 0, dispatcher idle, one `bash bin/mem search …` alive the whole time.
- Later, **15 orphaned `mem search` processes at once**, the oldest 35 minutes old. They accumulate at roughly 10/hour and survive the run that spawned them.
- The last thing in `run/logs/monolith.log` is the heartbeat, repeated forever:
  ```
  │ mem search: waiting for the model...
  │ mem search: waiting for the model...
  ```

Not a provider problem: while a search was hung, the same model answered a direct `bin/llm` call in 1–3s over the same pinned provider.

### Root cause

`bin/mem`, stage 2 of `action_search`:

```bash
local hb="${MEM_SEARCH_HEARTBEAT_S:-10}" _hb_pid=""
[[ "$hb" != 0 ]] && { ( while :; do sleep "$hb"; printf 'mem search: waiting for the model...\n' >&2; done ) & _hb_pid=$!; }
printf 'Search these memories for: %s\n\n%s' "$query" "$memories" | \
    llm ${SHELLM_FAST_MODEL:+-m "$SHELLM_FAST_MODEL"} -t "${MEM_SEARCH_MAX_TOKENS:-16000}" -s "..."
local rc=$?
if [[ -n "$_hb_pid" ]]; then kill "$_hb_pid" 2>/dev/null || true; ...
```

Two issues compound:

1. **The `llm` call has no upper bound.** `bin/llm`'s own curl guards cap a single attempt, but transient-failure retries mean total wall time is unbounded. Nothing here limits the whole stage.
2. **The heartbeat makes the hang invisible to the safety net that would otherwise end it.** The comment above it is explicit that it exists because "shellm kills a command that is silent for 30 seconds" — so by design it defeats `SHELLM_INACTIVITY_TIMEOUT`. When the model *does* answer in 31s that is correct; when the call never returns, the same mechanism guarantees the run is never killed.

[...]

PR #129, opened 2026-09-18
mem search: reap the heartbeat when the model call fails or is signalled

custos-1f916:
## What

The orphan half of #119. `action_search`'s stderr heartbeat was an infinite loop backgrounded from the search itself:

```bash
( while :; do sleep "$hb"; printf 'mem search: waiting for the model...\n' >&2; done ) & _hb_pid=$!
```

When the model call failed, `set -e` exited before the cleanup line; when the search was signalled, the `kill` never ran. Either way the loop was orphaned — reparented to init, still holding the write end of the `mem search | head` pipe, so the reader never saw EOF and hung. The bound half (the llm call itself could hang forever) was already fixed on main; this is the half that is still live.

## Fix

- Run the search in a subshell so its traps stay local to the call.
- Add EXIT/TERM/INT traps that reap the heartbeat.
- Make the heartbeat self-terminate: it polls `kill -0 "$search_pid"` and exits when the search dies, and its stdout is `/dev/null` so it never holds the result pipe.

The change is the same shape as the fix already in the deployed release (a1f58d25), extracted as a standalone hunk against main — no other part of the search rewrite.

## Tests

New `tests/test_mem_search_orphan.sh` (picked up by `tests/run-all.sh`, which CI runs):

1. a failed model call closes the pipeline and preserves the exit code (42)
2. a signalled search closes the pipeline and exits 143

Both hang (rc=124) on the old code; both pass on the fix. Existing `tests/test_mem_search.sh` passes unchanged (15/15), and the full mem sweep (`run-all.sh mem`) is green: 5/5 scripts.


nickjalbert:
LGTM thanks!

PR #131, opened 2026-09-19
Lift <tool_call>/<function=bash> markup into a bash fence

custos-1f916:
Qwen-family models sometimes answer in their native tool-call markup instead of a ```bash fence:

```
<tool_call>
<function=bash>
set -e … code …
</function>
</tool_call>
```

Run as-is, the first line is a bash syntax error and the whole turn is wasted — 4 of 46 responses on Custos, 2026-09-09, each one a lost wake.

**What this does:** when a reply has *no* ```bash fence but *does* contain `<tool_call>` / `<function=…>` / `<bash>` markup, `normalize_toolcall_markup` lifts the code between the openers and the first closer into a ```bash fence so it executes, and `extract_code` prepends a stderr notice telling the model to use a fence next time. A real fence always wins — markup words in a fenced reply are left alone.

Handles the four shapes observed in the wild: canonical `<tool_call><function=bash>…</function></tool_call>`, the hybrid `…</bash>`, a bare `<tool_call>` wrapper, and `<parameter=command>` inside `<function=bash>`.

**Tests:** `tests/test_extract_code_notice.sh` gains a section covering all four shapes (lifted, runs, carries the notice) plus the fence-wins case. 13 passed, 0 failed. `bash -n bin/shellm` clean.

Orthogonal to the no-fence notice already on main; co-occurs with it cleanly.


nickjalbert:
LGTM!  Thanks!

